PR #3, opened 2022-07-14
Move remainder of rust-rgb20 into rgb-std

cryptoquick:
So, watching you build RGB is like watching Leonardo Da Vinci paint the Mona Lisa. And I feel like this, opening this malformed PR:
Botched Ecce Homo

I was merely inspired by your original deprecated comment (before Wednesday's call), but alas, it's beyond my skill, and not terribly necessary, either.

That said, if you think there's some value in the work, I'm having trouble with resolving some of the `cargo check` errors.

If you want me to finish it, I could use a few pointers, and I can also squash the commits into a proper commit message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "amplify"
-version = "3.12.1"
+version = "3.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e45c604e86700ffea68a7d90c3ffb1508dad1120902f31dbb08afc942cf8acf3"
+checksum = "116019a174e912931d5b19ca7ab6a22596d12cdb1320358fad3368f0aba135a9"
 dependencies = [
  "amplify_apfloat",
  "amplify_derive",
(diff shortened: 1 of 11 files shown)


dr-orlovsky:
LOL regarding the pic :) No, you are overestimating my abilities.

On the topic: why do you think that having RGB20 in stdlib is a good idea? My point is to separate RGB stdlib and specific schema, such that new schema can be created without changing the stdlib at all - and in the future not even with the rust, but with AluVM/Contractum. If we put schema-specifics into the stdlib we would not be able to do that and will dim the lines of the abstraction and separation of concerns.

dr-orlovsky:
I think now I am getting it: I forgot to remove the deprecation warning from the RGB20 repo. It is undeprecated for at least a month for now and is activelly developed :D

cryptoquick:
Haha, yeah, that's what led me to do this work. It's probably a terrible idea, but I was also distracted by the existence of a `fungible` directory, and the moving of the Blind command. If rust-rgb20 is not to be deprecated, there's no point in finishing this work.

dr-orlovsky:
Sorry, it is my fault of not communicating/indicating properly

PR #7, opened 2022-10-25
Make bech32 encoding work for both types of consignments.

cryptoquick:
Currently, bech32 encoding (`.to_bech32_string()`) only works on contract consignments and not transfer consignments. This makes it work on both, which is needed for how we're implementing RGB on BitMask.

### DIFF
--- a/src/consignments/contract.rs
+++ b/src/consignments/contract.rs
@@ -14,18 +14,24 @@ use std::str::FromStr;
 
 use lnpbp_bech32::{FromBech32Str, ToBech32String};
 
-use super::Contract;
+use super::{ConsignmentType, InmemConsignment};
 
-impl lnpbp_bech32::Strategy for Contract {
+impl<T> lnpbp_bech32::Strategy for InmemConsignment<T>
+where T: ConsignmentType
+{
     const HRP: &'static str = "rgbc";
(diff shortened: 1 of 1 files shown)


dr-orlovsky:
Yeah, that was intentious. From what I learned, bech32  doesn't make sense for large amounts of data, where different encoding should be used. The only sense in bech32 is to transfer parts which can be typed by hand; for all other data something like BASE64, like in PGP, is more preferable.

While contract consignments are always just few KBs, the transfer ones will grow to multiple megabytes and gigabytes, since they include the history of the asset operations. Even after two or three transfers they do not fit a single telegram message.

cryptoquick:
Understood. I think we'll have to default to Carbonado encoding of consignments, then. It compresses and encrypts, so they can be stored as a file and easily transmitted out of band.

It might make sense to still make use of this PR for short consignments, for testing.

dr-orlovsky:
Where I can learn about it?

dr-orlovsky:
The plan is to move to Base64ASCIIarmoring in v0.10.0. Please feel free to re-open the issue if you think a different route should be taken.

Details:
- https://github.com/RGB-WG/rgb-node/discussions/243
- https://github.com/RGB-WG/rgb-std/issues/10

cryptoquick:
Okay, sounds good.

PR #13, opened 2022-11-20
Update dependencies

crisdut:


Cargo.toml
```diff
@@ -25,7 +25,7 @@ required-features = ["cli", "serde"]
 amplify = "3.13.0"
 lnpbp_bech32 = "~0.8.0"
 strict_encoding = { version = "~0.8.2", features = ["crypto", "chrono", "bitcoin"] }
-commit_verify = "~0.8.0"
+commit_verify = "0.8.1"
```
dr-orlovsky, line 28: I think we need to keep the `~`

dr-orlovsky (changes requested):
See comment. Can you also bump the version such that I can publish it after the merge?

Cargo.toml
crisdut: Yes, you are correct.

dr-orlovsky (approved):

PR #19, opened 2023-02-09
State reducers

dr-orlovsky:


dr-orlovsky:
Superseded due to v0.10 and new contract interfaces

PR #24, opened 2023-03-30
Fix RGB Invoice FromStr

crisdut:


dr-orlovsky (commented):
`+` was intentional: the beneficiary part can be a lightning node with address, which would contain `@`

crisdut:
> `+` was intentional: the beneficiary part can be a lightning node with address, which would contain `@`

Ok, sure.

I changed the `FromStr` implementation to use `+` instead of `@`.

dr-orlovsky (approved):
ACK https://github.com/RGB-WG/rgb-wallet/commit/7018ebee304d3cb9dfef401c934b9f42cb90a95b

PR #25, opened 2023-03-30
Fix TxID convertion

crisdut:
This PR fixies conversion between from bitcoin::TxId (rust-bitcoin) to bp::Txid

### DIFF
--- a/src/pay.rs
+++ b/src/pay.rs
@@ -21,7 +21,9 @@
 
 use std::collections::{BTreeMap, HashMap};
 use std::error::Error;
+use std::str::FromStr;
 
+use amplify::hex::ToHex;
 use bitcoin::hashes::Hash;
 use bitcoin::psbt::Psbt;
 use bp::seals::txout::CloseMethod;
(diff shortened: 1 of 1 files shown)


dr-orlovsky (commented):
But why do you think it was not working with the previous code? Since converting through string is much worse than just in-memory byte conversion: it is slower, does allocation + requires an expect.

dr-orlovsky:
1. No, bp-primitives doesn't construct hash
2. Even if it was display/from_str doesn't change anything in this regard

dr-orlovsky (changes requested):
Not needed

crisdut:
> Not needed

Sorry @dr-orlovsky,

I didn't understand what you meant.

I made this change because the current version returned an error when I tried the `rgb pay`.

```
[2023-03-31T17:44:19Z DEBUG rgb] Executing command: transfer
thread 'main' panicked at 'error paying invoice: InsufficientState', src/bin/rgb/command.rs:411:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This error occurs because when we convert `bitcoin::Txid` to `bp::Txid` using **to_byte_array** returns other information. Then, the system cannot get the `sum_inputs` correctly.

dr-orlovsky (changes requested):
Approach nACK: closing since it doesn't fix the source problem, which is in `bp-primitives` crate: a Txid must be displayed and read from string as a big-endian.

Fixed by me in https://github.com/BP-WG/bp-core/commit/50bf76d05be9bbaca2de00df59eae251ee336037

PR #31, opened 2023-04-06
RgbInvoice: change value from u64 to TypedState

zoedberg:
To support invoices without a specified RGB amount I was about to change the `value` field of the `RgbInvoice` struct from `u64` to an `Option<u64>`, but then I've noticed a TODO comment saying that `value` should be a `TypedState` field, so I've done that instead. I hope it's good enough. I've left a TODO comment (partially extracted from the removed one) for stuff that I don't see as urgent right now but eventually I'll open a PR for that too.

### DIFF
--- a/src/invoice.rs
+++ b/src/invoice.rs
@@ -116,19 +116,11 @@ impl std::fmt::Display for RgbInvoice {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         // TODO: Provide support for optionals & query
         let amt = self.value.to_string();
-        write!(
-            f,
-            "{}{}/{}/{}{}",
-            self.transport,
-            self.contract.to_baid58(),
-            self.iface,
-            if amt.is_empty() {
-                String::new()
-            } else {
-                format!("{amt}+")
-            },
-            self.beneficiary
-        )
+        write!(f, "{}{}/{}/", self.transport, self.contract.to_baid58(), self.iface)?;
+        if !amt.is_empty() {
+            write!(f, "{amt}+")?;
+        }
+        write!(f, "{}", self.beneficiary)
     }
 }
 
--- a/src/invoice.rs
+++ b/src/invoice.rs
@@ -22,11 +22,13 @@
 use std::num::ParseIntError;
 use std::str::FromStr;
 
+use baid58::ToBaid58;
 use bitcoin::{Address, Network};
 use bp::Chain;
 use fluent_uri::Uri;
 use indexmap::IndexMap;
 use rgb::{AttachId, ContractId, SecretSeal};
+use rgbstd::interface::TypedState;
 use strict_encoding::{InvalidIdent, TypeName};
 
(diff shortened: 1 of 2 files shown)


src/invoice.rs
```diff
@@ -112,6 +111,24 @@ pub enum InvoiceParseError {
...
+        write!(
+            f,
+            "{}{}/{}/{}{}",
+            self.transport,
+            self.contract,
+            self.iface,
+            match self.value {
+                TypedState::Amount(amt) => format!("{amt}+"),
+                _ => String::new(),
+            },
```
dr-orlovsky, line 126: We need to add `Dsiplay` impl for `TypedState` (if it's not there) and use it here.

src/invoice.rs
```diff
@@ -127,38 +144,40 @@ impl FromStr for RgbInvoice {
...
-                    (Err(_), Err(_)) => return Err(InvoiceParseError::Beneficiary(a.to_owned())),
-                    (Ok(_), Ok(_)) => panic!(
-                        "found a string which is both valid bitcoin address and UTXO blind seal"
-                    ),
-                };
-                (benefactor, value)
-            }
+        let (beneficiary_str, value) = match (assignment.next(), assignment.next()) {
+            (Some(a), Some(b)) => (b, TypedState::Amount(a.parse::<u64>()?)),
+            (Some(b), None) => (b, TypedState::Void),
```
dr-orlovsky, line 149: At least a TODO has to be added here to support other state types.

src/pay.rs
```diff
@@ -181,16 +182,25 @@ pub trait InventoryWallet: Inventory {
...
+                    }
+                    Ordering::Less => return Err(PayError::InsufficientState),
+                    Ordering::Equal => {}
+                }
+                main_builder
+                    .add_raw_state(assignment_id, beneficiary, TypedState::Amount(amt))?
+                    .complete_transition()?
+            }
+            _ => {
+                unimplemented!("only TypedState::Amount is currently supported")
```
dr-orlovsky, line 201:
```suggestion
                todo!("only TypedState::Amount is currently supported")
```

`unimplemented` in rust means "never going to be implemented". It is also not listed as TODOs by many IDEs.

dr-orlovsky (changes requested):
Thank you. Pls check my comments

src/invoice.rs
zoedberg: `Display` for `TypedState` is missing, yes. But I'm not sure how you want to represent values that are not of `TypedState::Amount` type, any suggestion?

dr-orlovsky: I did Display for TypedState. Just rebase on this commit: https://github.com/RGB-WG/rgb-wallet/pull/32/commits/ce92118465fcce28be1405776ac34a974c679786

dr-orlovsky:
Rebased on the latest changes and improved Display implementation

dr-orlovsky (approved):
ACK ef38257cad8b1fc61cbac845125f482058c07886

PR #33, opened 2023-04-12
RgbInvoice: make contract optional

zoedberg:
This PR makes the `contract` field of the `RgbInvoice` struct optional.
As discussed in a previous call this allows use cases where a user wants to receive an asset but doesn't care about which asset (e.g. collectibles donation).

### DIFF
--- a/src/invoice.rs
+++ b/src/invoice.rs
@@ -90,6 +90,10 @@ pub enum InvoiceParseError {
     /// invalid invoice.
     Invalid,
 
+    #[display(doc_comments)]
+    /// invalid contract ID.
+    InvalidContractId(String),
+
     #[from]
     Id(baid58::Baid58ParseError),
 
@@ -116,9 +120,9 @@ impl std::fmt::Display for RgbInvoice {
         let amt = self.owned_state.to_string();
         write!(f, "{}", self.transport)?;
         if let Some(contract) = self.contract {
-            write!(f, "{}/", contract.to_baid58())?;
+            write!(f, "{}", contract.to_baid58())?;
         }
-        write!(f, "{}/", self.iface)?;
+        write!(f, "/{}/", self.iface)?;
         if let Some(ref op) = self.operation {
             write!(f, "{op}/")?;
         }
@@ -160,12 +164,18 @@ impl FromStr for RgbInvoice {
 
         let mut next_path_index = 0;
 
-        let contract = match ContractId::from_str(&path[next_path_index]) {
+        let contract_id_str = &path[next_path_index];
+        let contract = match ContractId::from_str(contract_id_str) {
             Ok(cid) => {
                 next_path_index += 1;
                 Some(cid)
             }
-            Err(_e) => None,
+            Err(_e) => {
+                if !uri.path().to_string().starts_with('/') {
+                    return Err(InvoiceParseError::InvalidContractId(contract_id_str.clone()));
+                }
+                None
+            }
         };
 
         let iface = TypeName::try_from(path[next_path_index].clone())?;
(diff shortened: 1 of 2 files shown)


src/invoice.rs
```diff
@@ -154,7 +158,20 @@ impl FromStr for RgbInvoice {
...
-        let mut assignment = path[2].split('+');
+        let mut next_path_index = 0;
+
+        let contract = match ContractId::from_str(&path[next_path_index]) {
+            Ok(cid) => {
+                next_path_index += 1;
+                Some(cid)
+            }
+            Err(_e) => None,
+        };
```
dr-orlovsky, line 169: This opens vulnerability of introducing invalid contract names, which will be silently ignored.

I think what we need is to require that contract fragment must be always present, but may be empty meaning "any contract". In practice this should look like this: `rgb:/RGB20/6kzbKKffP6xftkxn9UP8gWqiC41W16wYKE5CYaVhmEve` (instead of `rgb:RGB20/6kzbKKffP6xftkxn9UP8gWqiC41W16wYKE5CYaVhmEve` form you are using currently in the Display and tests). With this, we skip contract if the first path segment is empty - but always fail with error if it is not containing a valid contract id.

dr-orlovsky (changes requested):
Please see my comment.

I think it is a good proposal and LGTM except one vulnerability, which I think is easy to address.

src/invoice.rs
zoedberg: I've tried adding a test invoice with invalid contract ID and, without changing the code, `from_str` already fails (with imprecise error message) because since `next_path_index` doesn't get increased it tries to parse the string `RGB20` as the beneficiary, thus it returns `Err(Beneficiary("RGB20"))`.

I'm ok with requiring the contract fragment to be always present, just note that
```rust
let path = uri
    .path()
    .segments()
    .map(|e| e.to_string())
    .collect::<Vec<String>>();
```
doesn't detect the empty fragment, so string `rgb:/RGB20/6kzbKKffP6xftkxn9UP8gWqiC41W16wYKE5CYaVhmEve` and `rgb:RGB20/6kzbKKffP6xftkxn9UP8gWqiC41W16wYKE5CYaVhmEve` are indistinguishable. Therefore, in order to detect invalid contract ID and provide a precise error message, I think we'll need to manually check the invoice string to see if after `{scheme}:` there is a slash or not, I'll update PR with a proposal fix

zoedberg: Pushed fix and updated tests

src/invoice.rs
```diff
@@ -160,12 +164,18 @@ impl FromStr for RgbInvoice {
...
-        let contract = match ContractId::from_str(&path[next_path_index]) {
+        let contract_id_str = &path[next_path_index];
+        let contract = match ContractId::from_str(contract_id_str) {
             Ok(cid) => {
                 next_path_index += 1;
                 Some(cid)
             }
-            Err(_e) => None,
+            Err(_e) => {
+                if !uri.path().to_string().starts_with('/') {
```
dr-orlovsky, line 174: not sure I get this part. Why do we need to check that the path doesn't start with `/`?

The rule that the first contract component must always be contract id. If it is not, this is `InvalidContractId` error. I think this `if` condition should be entirely omitted and we need just to error in this case.

zoedberg: This check is there to distinguish from a missing contract ID and an invalid one. As I explained in https://github.com/RGB-WG/rgb-wallet/pull/33#discussion_r1167467937 path segments don't include empty segments, that's the reason I introduced this check. Tests should confirm code is now working as expected

dr-orlovsky: Oh, github didn't show me your comment originally...

Ok, I see. However we need to structure this check differently in order to avoid unnecessary allocation. Proposing the change in my new review

src/invoice.rs
```diff
@@ -154,7 +162,26 @@ impl FromStr for RgbInvoice {
...
+                next_path_index += 1;
+                Some(cid)
+            }
+            Err(_e) => {
+                if !uri.path().to_string().starts_with('/') {
+                    return Err(InvoiceParseError::InvalidContractId(contract_id_str.clone()));
+                }
+                None
+            }
+        };
```
dr-orlovsky, line 179:
```suggestion
        let contract = match ContractId::from_str(contract_id_str) {
            Ok(cid) => {
                next_path_index += 1;
                Some(cid)
            }
            Err(_) if !uri.path().as_str().starts_with('/') =>  {
                return Err(InvoiceParseError::InvalidContractId(contract_id_str.clone()));
            }
            Err(_) => None,
        };
```

dr-orlovsky (changes requested):

dr-orlovsky (approved):
ACK https://github.com/RGB-WG/rgb-wallet/commit/2442a7d806aab4fac3645db526f764e394a24749

Thank you for going through all these revisions!

PR #51, opened 2023-05-17
Add support to RGB21 (ContractBuilder)

crisdut:


std/src/interface/builder.rs
```diff
@@ -379,6 +392,43 @@ impl<Seal: ExposedSeal> OperationBuilder<Seal> {
...
     }
 
+    pub fn add_data_state(
+        mut self,
+        name: impl Into<FieldName>,
+        seal: impl Into<BuilderSeal<Seal>>,
+        value: impl StrictSerialize,
+    ) -> Result<Self, BuilderError> {
+        let name = name.into();
+        let serialized = value.to_strict_serialized::<65535>()?;
```
dr-orlovsky, line 402:
```suggestion
        let serialized = value.to_strict_serialized::<U16>()?;
```

And add `use amplify::confinement::U16`

dr-orlovsky (commented):
Once small nit to ensure we use correct value in serialization generic

PR #65, opened 2023-06-14
Add OutpointFilter to filter wallet-based outputs

dr-orlovsky:
This should significantly simplify wallet APIs such that contract interfaces will now return only a state which is specific to some wallet descriptor.

### DIFF
--- a/std/src/interface/contract.rs
+++ b/std/src/interface/contract.rs
@@ -19,6 +19,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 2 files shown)


dr-orlovsky:
NB: this does not (yet) provides guarantees that the issuer does not lie about the amount of the issued asset.

To ensure that the issuer can't fake `issuedSupply` value in genesis we need an AluVM script checking the sum of pedersen commitments of the spent assets. I am working on adding it (but at least you can start playing with the new API since the API wouldn't change)

PR #81, opened 2023-07-06
Add blank state transitions to state

### LINKED ISSUE #80: Blank state transitions do not update to state data
Using our rgb-sandbox I tried to issue 2 assets to the same UTXO, then transfer the first asset to another wallet and then check the the second asset's allocation, which I expect to be on the change outpout at that point.

What I find is that, after the transfer has completed, the second asset's allocation is still on the original issuance UTXO, which has now been spent.

To give an example, this is the situation after issuing the two assets:
- wallet unspents:
 ```
    8a675edd1a228bdbe24254bf1675a057275b49ade7bac9049513a77a40c6f59b:0
```
- 1st asset allocations
```
    amount=2000  utxo=8a675edd1a228bdbe24254bf1675a057275b49ade7bac9049513a77a40c6f59b:0
```
- 2nd asset allocations
```
    amount=2000  utxo=8a675edd1a228bdbe24254bf1675a057275b49ade7bac9049513a77a40c6f59b:0
```

Then a transfer (txid `296ab67bd423551c87229e9371b85ee5c62b60417d8015bed146d49d71bb784f`) sends 100 units of the first asset to another wallet and this is the resulting situation (on the issuer/sender side):
- wallet unspents:
 ```
    296ab67bd423551c87229e9371b85ee5c62b60417d8015bed146d49d71bb784f:1
```
- 1st asset allocations
```
    amount=1900  utxo=296ab67bd423551c87229e9371b85ee5c62b60417d8015bed146d49d71bb784f:1
```
- 2nd asset allocations
```
    amount=2000  utxo=8a675edd1a228bdbe24254bf1675a057275b49ade7bac9049513a77a40c6f59b:0
```

This issue can be reproduced using [this branch](https://github.com/nicbus/rgb-sandbox/tree/v0.10_blank) of rgb-sandbox.

dr-orlovsky:
Fixes #80

The bug was caused by the fact that we do state update only on consignment consumation - and not on production of state transitions. This PR fixes that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -700,7 +700,7 @@ dependencies = [
 [[package]]
 name = "rgb-core"
 version = "0.10.4"
-source = "git+https://github.com/RGB-WG/rgb-core#1de67f7bcd655bbb497074da50c0872cb1ae433a"
+source = "git+https://github.com/RGB-WG/rgb-core?branch=preserve-historic-state#23706b572aa4e6acf371b7dcb9c15b82e56c0dce"
 dependencies = [
  "aluvm",
  "amplify",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,4 +78,4 @@ wasm-bindgen-test = "0.3"
 features = [ "all" ]
 
 [patch.crates-io]
-rgb-core = { git = "https://github.com/RGB-WG/rgb-core" }
+rgb-core = { git = "https://github.com/RGB-WG/rgb-core", branch = "preserve-historic-state" }
--- a/std/src/interface/builder.rs
+++ b/std/src/interface/builder.rs
@@ -357,7 +357,13 @@ impl<Seal: ExposedSeal> OperationBuilder<Seal> {
         let serialized = value.to_strict_serialized::<{ u16::MAX as usize }>()?;
 
         // Check value matches type requirements
-        let Some(type_id) = self.iimpl.global_state.iter().find(|t| t.name == name).map(|t| t.id) else {
+        let Some(type_id) = self
+            .iimpl
+            .global_state
+            .iter()
+            .find(|t| t.name == name)
+            .map(|t| t.id)
+        else {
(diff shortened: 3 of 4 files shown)


zoedberg:
I've tried this fix on rgb-lib. It solves the blank transition issue, but it breaks some other tests that were previously succeeding. We'll try to reproduce the new bugs with the sandbox so that you can see it yourself

nicbus:
Since an rgb-lib test fails when the code from this PR is used, using our sandbox I tried to initiate a transfer (up to `rgb transfer`) but then abort it (don't validate, broadcast the tx nor accept the transfer on either side), then re-start the transfer from scratch (including re-generating the recipient invoice).

Once the first transfer has been initiated, `rgb state` shows the change allocation from that transfer (instead of the original one), although the related tx was never broadcast.

This situation can be reproduced using [this branch](https://github.com/nicbus/rgb-sandbox/tree/v0.10_aborted_transfer) of the sandbox.

As an example, given the following initial situation:
- outpoints:
```
dbfa837f699077039b395a37277b4d160f91c422184714554aeed204f27842b5:0
```
- allocations:
```
2000  utxo=dbfa837f699077039b395a37277b4d160f91c422184714554aeed204f27842b5:0
```
after initiating a transfer (txid `946a3a73598a9ec45033f354596665fd0b3f070ced821a0492a9636c5a15d2b2`) this is what I get:
- outpoints:
```
dbfa837f699077039b395a37277b4d160f91c422184714554aeed204f27842b5:0
```
- allocations:
```
amount=1900  utxo=946a3a73598a9ec45033f354596665fd0b3f070ced821a0492a9636c5a15d2b2:0
```

#82 might also be related.

dr-orlovsky:
@nicbus I fixed the problem of removing historic state in https://github.com/RGB-WG/rgb-core/pull/166

Please re-test with that fix - I think all tests will be passing now

PR #95, opened 2023-07-25
Fix assigning states to default velocity if no outputs matching velocity are given

dr-orlovsky:


nicbus:
checked with `rgb-sandbox` updating `rgb-wallet` to commit `2f272b0` and the transfer now succeeds

the demo then fails because it can't detect balances correctly, but that's [another issue](https://github.com/RGB-WG/rgb-wallet/issues/98)

so I confirm the issue at hand is solved

PR #102, opened 2023-08-28
fix receiving asset a second time from a different consignment "erases" old one

zoedberg:
This is a fix proposal for https://github.com/RGB-WG/rgb-wallet/issues/101.

IMO it's safe to never empty out the `contract_index`. Running all our available tests with this fix I haven't found any regressions but I may be missing something.


std/src/persistence/stock.rs
```diff
@@ -173,10 +173,15 @@ impl Stock {
...
-        })?;
+        match self.contract_index.get(&contract_id) {
+            None => {
+                self.contract_index.insert(contract_id, ContractIndex {
+                    public_opouts: empty!(),
+                    outpoint_opouts: empty!(),
+                })?;
+            }
+            Some(_) => {}
+        }
```
dr-orlovsky, line 184: In rust it will be more idiomatic to have
```suggestion
        self.contract_index.entry(contract_id).or_default();
```
it will do the same thing

dr-orlovsky (approved):
Yes, I think you are right in spotting the exact bug. Pls see my comment on how we can do in just one line.

std/src/persistence/stock.rs
zoedberg: I tried to apply the suggested change but then it fails building with this error:
```
59  | pub struct ContractIndex {
    | ------------------------ doesn't satisfy `ContractIndex: std::default::Default`
...
176 |         self.contract_index.entry(contract_id).or_default();
    |                                                ^^^^^^^^^^ method cannot be called on `Entry<'_, ContractId, ContractIndex>` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `ContractIndex: std::default::Default`
help: consider annotating `ContractIndex` with `#[derive(Default)]`
    |
59  + #[derive(Default)]
60  | pub struct ContractIndex {
    |
```
and by annotating `ContractIndex` with `#[derive(Default)]` or manually implementing the `Default` trait I receive the following error:
```
error[E0119]: conflicting implementations of trait `StrictDumb` for type `ContractIndex`
  --> /rgb-wallet/std/src/persistence/stock.rs:57:22
   |
57 | #[derive(StrictType, StrictDumb, StrictEncode, StrictDecode)]
   |                      ^^^^^^^^^^
   |
   = note: conflicting implementation in crate `strict_encoding`:
           - impl<T> StrictDumb for T
             where T: StrictType, T: std::default::Default;
   = note: this error originates in the derive macro `StrictDumb` (in Nightly builds, run with -Z macro-backtrace for more info)
```

dr-orlovsky: Yes, `Entry` API doens't (yet) work with confined collections from `amplify`, that was my bad.

I did it in a slightly different way - see my changes in the pushed commits

zoedberg: I've checked, changes work, thanks

PR #104, opened 2023-09-18
Handle chain information in explicit form in invoices

### LINKED ISSUE #103: Network not set on invoice
Despite setting the network argument, we get still invoices like this:
`rgb:Cbw1h3zbHgRhA6sxb4FS3Z7GTpdj9MLb7Do88qh5TUH1/RGB20/1+utxob02FKN7XB4N5LWc1d6cXfkByZ8d1jB4BZhee1vdU7MwffTtLkiK3`
When parsed, it lacks the network argument.
This will be helpful for users to better verify and troubleshoot their transfers.

dr-orlovsky:
This is a breaking change since inside the invoice structure we make `chain` non-optional; plus we add new error cases and make `set_chain` return `Result`. Thus I am making this PR against `v11` branch.

Closes #103

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,7 +260,7 @@ dependencies = [
 [[package]]
 name = "bp-primitives"
 version = "0.10.7"
-source = "git+https://github.com/BP-WG/bp-core?branch=primitives#b0c8538c223c29e85ad58fe4f6be59d6c8555515"
+source = "git+https://github.com/BP-WG/bp-core?branch=primitives#a24e6ffca3fce442706f2cab59613d8d66dc7c20"
 dependencies = [
  "amplify",
  "commit_verify",
--- a/invoice/src/builder.rs
+++ b/invoice/src/builder.rs
@@ -32,15 +32,16 @@ pub struct RgbInvoiceBuilder(RgbInvoice);
 
 impl RgbInvoiceBuilder {
     pub fn new(beneficiary: impl Into<Beneficiary>) -> Self {
+        let beneficiary = beneficiary.into();
         Self(RgbInvoice {
             transports: vec![RgbTransport::UnspecifiedMeans],
             contract: None,
             iface: None,
             operation: None,
             assignment: None,
-            beneficiary: beneficiary.into(),
+            chain: beneficiary.chain_info().unwrap_or(Chain::Bitcoin),
+            beneficiary,
             owned_state: TypedState::Void,
-            chain: None,
             expiry: None,
             unknown_query: none!(),
         })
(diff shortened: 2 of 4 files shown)


cryptoquick (approved):
ACK
Very comprehensive tests cases. Looks good.
I also agree, 0.11 is a good place to put this.

fedsten:
Any reason why it is non-optional instead of optional as discussed in #103?

dr-orlovsky:
@fedsten the parameter is optional as was discussed, thus the invoice string is backwards-compatible. However, for each given invoice we always know which network it belongs to (since we are defaulting to mainnet as discussed), thus we must have a non-optional field.

fedsten:
Got it,  for clarifying.

dr-orlovsky:
In other words, a wallet will always know and be sure to which network a given invoice belongs. If the invoice has conflicting network information (like an address network doesn't match the network provided in a parameter) it will error during the parse procedure. But each parsed invoice always knows which network it is valid for,

zoedberg (changes requested):
I'm trying to review the PR but executing tests I'm encountering some issues.
It seems that the `LoadError` is defined only when the `fs` feature is active but then used in other parts of code available without that feature. Moreover, if I run tests activating all features, I receive the following error:
```
error[E0599]: no method named `assume_checked` found for struct `bp::Address` in the current scope
   --> invoice/src/parse.rs:454:14
    |
452 |           let addr_bc = Address::from_str("bc1qpws79r3ea4yy2ujsahwrmy2gutdj8w5whnhket")
    |  _______________________-
453 | |             .unwrap()
454 | |             .assume_checked();
    | |             -^^^^^^^^^^^^^^ method not found in `Address`
    | |_____________|
    | 

error[E0599]: no method named `assume_checked` found for struct `bp::Address` in the current scope
   --> invoice/src/parse.rs:457:14
    |
455 |           let addr_tc = Address::from_str("mxVFsFW5N4mu1HPkxPttorvocvzeZ7KZyk")
    |  _______________________-
456 | |             .unwrap()
457 | |             .assume_checked();
    | |             -^^^^^^^^^^^^^^ method not found in `Address`
    | |_____________|
    | 

For more information about this error, try `rustc --explain E0599`.
error: could not compile `rgb-invoice` (lib test) due to 2 previous errors
```

dr-orlovsky:
@zoedberg wait, we do not have CI set up for this repo? Oh, I'll fix it.

I did this PR originally against v0.10 branch which was used rust-bitcoin v0.30. Then I understood it's breaking, so moved to v0.11, which doesn't depend on rust-bitcoin at all and have a different `Address` type (from `bp-std`) with different API. Thus the issue. Will fix.

dr-orlovsky:
Closing in favor of solution from https://github.com/RGB-WG/rgb-std/pull/118

PR #112, opened 2023-12-12
ifaces: use correct fungible builder type

TheWaWaR:
Since `Terminal.seals` is a `SmallXxx` collection, so the seal quantity may reach up to 65535, then the corresponding builder should allow seal quantity up to 65535.

```rust
pub struct Terminal {
    pub seals: SmallOrdSet<TerminalSeal>,
    pub tx: Option<Tx>,
}
```

It's not a protocol change, if you wish I can port this fix to `v0.10` branch.

### DIFF
--- a/src/interface/builder.rs
+++ b/src/interface/builder.rs
@@ -357,10 +357,10 @@ pub struct OperationBuilder<Seal: ExposedSeal> {
     asset_tags: TinyOrdMap<AssignmentType, AssetTag>,
 
     global: GlobalState,
-    // rights: TinyOrdMap<AssignmentType, Confined<HashSet<BuilderSeal<Seal>>, 1, U8>>,
+    // rights: TinyOrdMap<AssignmentType, Confined<HashSet<BuilderSeal<Seal>>, 1, U16>>,
     fungible:
-        TinyOrdMap<AssignmentType, Confined<HashMap<BuilderSeal<Seal>, RevealedValue>, 1, U8>>,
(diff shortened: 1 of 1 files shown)


dr-orlovsky:
You are right that we should allow up to 2^16 assignments in the operation builder, but this is not related to terminal seals, and is instead related to the number of assignments inside the genesis/state transition, which is 2^16.

Terminal seals can be collected from multiple state transition, thus limit there is unrelated to this one.

Anyway, thank you for spotting that. Can you please also update `data`, `rights` and `attach` in the same way?

Thank you for finding this!

TheWaWaR:
I can only find `rights`(commented)/`fungible`/`data`, but not `attach`.
```rust
pub struct OperationBuilder<Seal: ExposedSeal> {
    // TODO: use references instead of owned values
    schema: SubSchema,
    iface: Iface,
    iimpl: IfaceImpl,
    asset_tags: TinyOrdMap<AssignmentType, AssetTag>,

    global: GlobalState,
    // rights: TinyOrdMap<AssignmentType, Confined<HashSet<BuilderSeal<Seal>>, 1, U8>>,
    fungible:
        TinyOrdMap<AssignmentType, Confined<HashMap<BuilderSeal<Seal>, RevealedValue>, 1, U8>>,
    data: TinyOrdMap<AssignmentType, Confined<HashMap<BuilderSeal<Seal>, RevealedData>, 1, U8>>,
    // TODO: add attachments
    // TODO: add valencies
}
```

dr-orlovsky (approved):
ACK https://github.com/RGB-WG/rgb-std/commit/d4284662b13e0dca867e179dc1076309b93cfba5

Thank you!

PR #113, opened 2023-12-15
Changes required by RGB wallet PSBT operations

dr-orlovsky:
Also includes support for combined tapret/opret commitments.

Mostly ready - will mark for review the moment wallet library will produce PSBTs.


src/persistence/inventory.rs
```diff
@@ -301,22 +344,39 @@ pub trait Inventory: Deref<Target = Self::Stash> {
...
+    }
+
+    #[doc(hidden)]
+    unsafe fn consume_anchor(
         &mut self,
-        anchor: Anchor<mpc::MerkleBlock>,
+        anchor: XAnchor<mpc::MerkleBlock>,
     ) -> Result<(), InventoryError<Self::Error>>;
 
-    /// # Safety
```
cryptoquick, line 314: Can you re-add the docs for the unsafe code? Just indicating why unsafe and under what conditions

src/persistence/stock.rs
```diff
@@ -471,20 +462,19 @@ impl Inventory for Stock {
         self.consume_consignment(transfer, resolver, force)
     }
 
-    fn consume_anchor(
+    unsafe fn consume_anchor(
```
cryptoquick, line 465: Why unsafe?

cryptoquick (commented):
That was a lot of work! Just a couple of concerns... The unsafes make me nervous. If they're necessary, then maybe put some time into documenting why they need to be unsafe.

invoice/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[allow(clippy::module_inception)]
+mod invoice;
+mod parse;
+mod builder;
+mod amount;
+
+pub use amount::{Amount, CoinAmount, Precision};
+pub use builder::RgbInvoiceBuilder;
+pub use invoice::{Beneficiary, InvoiceState, RgbInvoice, RgbTransport};
```
crisdut, line 40: For avoid compile error: `ambiguous because of multiple potential`.

```suggestion
pub use crate::invoice::{Beneficiary, InvoiceState, RgbInvoice, RgbTransport};
```

invoice/src/amount.rs
```diff
@@ -0,0 +1,303 @@
...
+            fract,
+            precision,
+        }
+    }
+}
+
+impl Display for CoinAmount {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let mut int = self.int.to_string();
+        if f.alternate() {
```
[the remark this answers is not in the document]
dr-orlovsky: Unsafe is required because of `from_utf8_unchecked` - since we are sure the string is UTF8

Chunks of three bytes to print the separators: `100'000'000`

invoice/src/builder.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use super::{Beneficiary, InvoiceState, Precision, RgbInvoice, RgbTransport, TransportParseError};
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct RgbInvoiceBuilder(RgbInvoice);
+
+#[allow(clippy::result_large_err)]
+impl RgbInvoiceBuilder {
+    pub fn new(beneficiary: impl Into<Beneficiary>) -> Self {
+        Self(RgbInvoice {
```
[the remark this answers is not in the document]
dr-orlovsky: With `Default` you will get an invalid invoice: no transports, no beneficiary; and this will require converting `finish` to return `Result` type (and introducing another error type). I am not sure this is better.

src/persistence/stock.rs
dr-orlovsky: To prevent it from being called from any place other than the clearly defined single place.

Not a best practice, I know - but it is easier not to mistake/miss invalid calls in this way

src/persistence/inventory.rs
dr-orlovsky: I removed it so it since it will not be seen anyway due to the added `doc(hidden)` method... I do not know how to hide such things properly and at the same time provide those implementing the trait with enough info...

PR #116, opened 2023-12-26
Fix bundle merge reveal check

crisdut:
**Description**

This PR intent fix the merge reveal bundle check. Also, adjusts the error message for correct naming.

### DIFF
--- a/src/accessors/merge_reveal.rs
+++ b/src/accessors/merge_reveal.rs
@@ -42,8 +42,8 @@ pub enum MergeRevealError {
     #[display(inner)]
     AnchorMismatch(MergeError),
 
-    /// the merged bundles contain excessive transactions.
-    ExcessiveTransactions,
+    /// the merged bundles contain excessive transitions.
+    ExcessiveTransitions,
 
     /// contract id provided for the merge-reveal operation doesn't matches
     /// multi-protocol commitment.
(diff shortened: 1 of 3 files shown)


src/accessors/merge_reveal.rs
```diff
@@ -187,12 +187,12 @@ impl<Seal: ExposedSeal> MergeReveal for Assignments<Seal> {
 impl MergeReveal for TransitionBundle {
     fn merge_reveal(mut self, other: Self) -> Result<Self, MergeRevealError> {
         debug_assert_eq!(self.commitment_id(), other.commitment_id());
-        if self.known_transitions.len() + other.known_transitions.len() > self.input_map.len() ||
+        if self.input_map.len() < other.known_transitions.len() ||
```
dr-orlovsky, line 190: I think we need to take into account both `self` and `other` when do comparison. Why do you think the original version was buggy?

dr-orlovsky: All right, I see. I think it should be this way: first we do `extend`, than we do check

crisdut: I imagined that the original version made a purposeful comparison between `self` and `other`, so I adjusted it to a verification that made more sense.

So, should we first do the `extend` and then check if the amount of `input_map` is less than the `known_transitions` ?

dr-orlovsky: Yes

crisdut: Done

dr-orlovsky (approved):
ACK https://github.com/RGB-WG/rgb-std/commit/d53454a5c6c440b163b7eb399aa276049dc6a08a

PR #126, opened 2024-01-07
Fungible operation history

dr-orlovsky:
Prerequisites for https://github.com/RGB-WG/rgb/pull/92. Include:
- new concept of witness filters;
- new concept of state change types which help to track past operation effects
- `AmountChange`: a case of state change type which tracks the change in the balance as an effect of past fungible transfers
- refactored typed states
- `ContractIface::fungible_ops` command reconstructing history of operations affecting some fungible state (with changes expressed as AmountChange)
- `Rgb20::transfer_history` wrapper which simplifies the call for RGB20-interfaces contracts

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -655,7 +655,7 @@ dependencies = [
 [[package]]
 name = "rgb-core"
 version = "0.11.0-beta.4"
-source = "git+https://github.com/RGB-WG/rgb-core?branch=v0.11#b1232f5f77ff253bbe79bdcee04a2cf0d4ef0c05"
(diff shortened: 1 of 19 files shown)


crisdut:
Really good @dr-orlovsky,

When I implemented a similar history feature in the bitmask wallet, I encountered a scenario where we needed hidden "invalid allocations" in stash history. This scenario occurs because we have implemented RBF,  when the sender generate new bitcoin transaction using RBF, the previous transitions in RGB stash turns invalid.

Did you test this scenario? 

It would be interesting to have this option to hide "invalid allocations" by default.

dr-orlovsky:
This works/is blocked by using appropriate witness filter. You can even work with block reorgs and lightning channel history by using them. 

I will expose witness filter customization to outside of Runtime

crisdut (approved):

PR #127, opened 2024-01-07
Fix blank transition builder

crisdut:
Alternative to https://github.com/RGB-WG/rgb-std/pull/108

Close #81 
Close #108

### DIFF
--- a/src/interface/rgb21.rs
+++ b/src/interface/rgb21.rs
@@ -101,6 +101,10 @@ impl Allocation {
     pub fn with(index: TokenIndex, fraction: OwnedFraction) -> Allocation {
         Allocation(index, fraction)
     }
+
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
In https://github.com/RGB-WG/rgb-std/pull/126 I moved from using dedicated `*Allocation` type and instead re-use types from RGB Core this way: https://github.com/RGB-WG/rgb-std/pull/126/files#diff-16373c59058401969ad1892e8d953d68a88387b2b435267816c230033c083201R83-R87

Can you please re-base this PR on top of that one to make them matching?

crisdut:
@dr-orlovsky, 

Done.

dr-orlovsky (approved):
ACK 9e7b89da4bc8968cad1273ff17b72f4744b0ab54

PR #132, opened 2024-01-16
fix blank_builder to avoid AssetTagMissed error

zoedberg:
When sending an asset allocated to an UTXO that has also other assets allocated we need to construct the blank transitions (using the `blank_builder` method).
But when doing
```rust
blank_builder = blank_builder
    .add_input(opout, state.clone())?
    .add_owned_state_raw(opout.ty, seal, state)?;
```
the call to `add_owned_state_raw` raises the `AssetTagMissed` error (from the internal call to the `asset_tag` method), because `self.asset_tags` is empty.
With the proposed fix this error doesn't arise anymore. The fix consists in calling `add_asset_tag_raw` the same way it's currently done in the `transition_builder` method.

### DIFF
--- a/src/persistence/inventory.rs
+++ b/src/persistence/inventory.rs
@@ -514,7 +514,7 @@ pub trait Inventory: Deref<Target = Self::Stash> {
             )));
         }
 
-        let builder = if let Some(iimpl) = schema_ifaces.iimpls.get(&iface.iface_id()) {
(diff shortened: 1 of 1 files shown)


dr-orlovsky (approved):
ACK 5ed49570b69d113e0ea23132c17dc2a3cc1e9a99

Very good, thank you!

PR #136, opened 2024-01-22
fix TransitionBundle merge_reveal

zoedberg:
This PR fixes https://github.com/RGB-WG/rgb/issues/101

The bug was that the method `merge_reveal` (implemented for the `TransitionBundle`) was calling `extend` assuming that `other` had more revealed data than `self`. This is not always the case, so the updated version merges the structures keeping the most revealed condition among the two.

### DIFF
--- a/src/accessors/merge_reveal.rs
+++ b/src/accessors/merge_reveal.rs
@@ -188,12 +188,19 @@ impl MergeReveal for TransitionBundle {
     fn merge_reveal(mut self, other: Self) -> Result<Self, MergeRevealError> {
         debug_assert_eq!(self.commitment_id(), other.commitment_id());
 
-        if self
-            .known_transitions
-            .extend(other.known_transitions)
-            .is_err() ||
-            self.input_map.len() > self.known_transitions.len()
-        {
+        for (opid, transition) in self.known_transitions.keyed_values_mut() {
+            if let Some((_, other_transition)) =
+                other.known_transitions.iter().find(|(o, _)| o == &opid)
+            {
+                transition.assignments = transition
+                    .assignments
+                    .clone()
+                    .merge_reveal(other_transition.assignments.clone())
+                    .unwrap();
+            }
+        }
+
+        if self.input_map.len() > self.known_transitions.len() {
             return Err(MergeRevealError::ExcessiveTransitions);
         }
         Ok(self)
--- a/src/accessors/merge_reveal.rs
+++ b/src/accessors/merge_reveal.rs
@@ -188,17 +188,14 @@ impl MergeReveal for TransitionBundle {
     fn merge_reveal(mut self, other: Self) -> Result<Self, MergeRevealError> {
         debug_assert_eq!(self.commitment_id(), other.commitment_id());
 
-        for (opid, transition) in self.known_transitions.keyed_values_mut() {
-            if let Some((_, other_transition)) =
-                other.known_transitions.iter().find(|(o, _)| o == &opid)
-            {
-                transition.assignments = transition
-                    .assignments
-                    .clone()
-                    .merge_reveal(other_transition.assignments.clone())
-                    .unwrap();
+        let mut self_transitions = self.known_transitions.into_inner();
+        for (opid, other_transition) in other.known_transitions {
+            if let Some(mut transition) = self_transitions.remove(&opid) {
+                transition = transition.merge_reveal(other_transition)?;
+                self_transitions.insert(opid, transition);
             }
         }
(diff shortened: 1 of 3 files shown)


src/accessors/merge_reveal.rs
```diff
@@ -188,12 +188,19 @@ impl MergeReveal for TransitionBundle {
...
 
-        if self
-            .known_transitions
-            .extend(other.known_transitions)
-            .is_err() ||
-            self.input_map.len() > self.known_transitions.len()
-        {
+        for (opid, transition) in self.known_transitions.keyed_values_mut() {
+            if let Some((_, other_transition)) =
+                other.known_transitions.iter().find(|(o, _)| o == &opid)
```
dr-orlovsky, line 193: We can use `other.known_transitions.remove(opid)` and by doing that avoid clone below

dr-orlovsky (changes requested):
Just one change, and you did a great job figuring the source of the problem!

src/accessors/merge_reveal.rs
zoedberg: `other` is not mutable so I don't think calling `remove` is possible (but it was a good point, thanks)

dr-orlovsky: Other is here by value, meaning you can just add `mut` in front of it in the method signature

zoedberg: You're right. I've updated the PR, thanks again

zoedberg: Actually, I've tried this change on rgb-lib and it seems that now another test (where a wallet sends back what it has received) fails. So I'm going to revert this change.

zoedberg: For more context, using `remove` instead of `find` makes the receiver fail the command `consignment.validate` with the error `consignment is invalid and can't be imported.` when the command is performed a second time (I guess because `remove` is removing data from `self.anchors`).

zoedberg: I'm not sure what you intend. `other` comes from the `consume_bundle` method in the`hoard` module:
```rust
pub fn consume_bundle(&mut self, bundle: TransitionBundle) -> Result<(), ConsumeError> {
    let bundle_id = bundle.bundle_id();
    match self.bundles.get_mut(&bundle_id) {
        Some(b) => *b = b.clone().merge_reveal(bundle)?,
        None => {
            self.bundles.insert(bundle_id, bundle)?;
        }
    }
    Ok(())
}
```
so by changing the signature of `merge_reveal` like so:
```rust
fn merge_reveal(self, mut other: Self) -> Result<Self, MergeRevealError> {
```
and calling `remove` we are changing `self.bundles`,  causing so the previously mentioned bug. Not sure why we should remove and then re-insert but maybe I'm missing something

dr-orlovsky: We remove from `other`, not `self`: `other.known_transitions.find` -> `other.known_transitions.remove`

dr-orlovsky: Ok, I got it, sorry. `transitions` come from `self` and not `other`.

dr-orlovsky: But you can still do `other.transition.assignments.merge_reveal(self...)` instead of `self.transition.assignments.merge_reveal(other...)` and at least save one clone.

Sorry, these clones are really very performance-affecting and it is nice to avoid them wherever is possible. Especially here, where the operation is `O(N)` with `N` is the number of transitions in a stash!

Self = stash
other = consignment

zoedberg: Not sure I've done what you're asking, but I tried doing:
```rust
transition.assignments = other_transition
    .assignments
    .clone()
    .merge_reveal(transition.assignments)
    .unwrap();
```
but then I receive the error
```
move occurs because `transition.assignments` has type `Assignments<BlindSeal<TxPtr>>, which does not implement the `Copy` trait
```
And if I try removing the previous clone (the one on `other.transition.assignments`) I receive the error
```
move occurs because `other_transition.assignments` has type `Assignments<BlindSeal<TxPtr>>`, which does not implement the `Copy` trait
```

dr-orlovsky: I added commit showing what I have meant :)

zoedberg: Great, thanks

dr-orlovsky: Can I merge this? Can you pls confirm that it works with your tests?

zoedberg: I confirm it works. You can merge this, thank you

PR #142, opened 2024-02-01
convert OperationBuilder map fields to BTreeMap

zoedberg:
This PR changes the `OperationBuilder` map fields from `HashMap`s to `BTreeMap`s, allowing to construct deterministic transitions (because `BTreeMap` is an ordered map, while with the `HashMap` we cannot guarantee the inserted pairs are always in the same order)


dr-orlovsky (approved):
ACK 8c50ba5f112f0350c414455e870a5a2626452d28

Good point

PR #147, opened 2024-02-15
Fix transferID generation

crisdut:
**Description**

This PR corrects the creation of the transfer ID, using the audited version of commitments.

PS: I'm pointing to branch v0.11, because I noticed that you had already started the work of using the latest version of rgb-core.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,7 +643,7 @@ dependencies = [
 [[package]]
 name = "rgb-core"
 version = "0.11.0-beta.4"
-source = "git+https://github.com/RGB-WG/rgb-core?branch=vesper#6165362bb723c72f400ff0a2ff9cd7429f6751f2"
+source = "git+https://github.com/RGB-WG/rgb-core?branch=vesper#1759eab75a1d2f94f402db6897bc76abc1221441"
 dependencies = [
  "aluvm",
  "amplify",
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
Sorry, I am not getting why do we need these changes

crisdut:
> Sorry, I am not getting why do we need these changes

Okay, firstly, I updated the version of rgb-core containing its latest fixes.

I ran the latest versions of all packages in my test environment.

Then I noticed two problems:

1. This [line](https://github.com/RGB-WG/rgb-std/pull/147/files#diff-e0c4aa2cc688804319d5e7a8f1c5ec13b8b6eb4af8d4ffdec3963a23347fdae6L81) realize recursively calls of the `Transfer::commit_encode` method until the stackoverflow panics occurs.
2. A create the wrapper type [SecretSealSet](https://github.com/RGB-WG/rgb-std/pull/147/files#diff-83a44a7fd9e1403b2135b1553467b3fb480951f948fa64dff9089992a597f833R192-R218) to avoid occurs the error `commit encoder can commit only to named types`, similar to what happened [here](https://github.com/RGB-WG/rgb-core/pull/204#issuecomment-1946271249).

dr-orlovsky:
I see. It seems we need to refactor the whole way how this works. My variant is in https://github.com/RGB-WG/rgb-std/pull/148/commits/251cecf908f61a92df534879da1c693cd3c12cb9

crisdut:
> I see. It seems we need to refactor the whole way how this works. My variant is in [251cecf](https://github.com/RGB-WG/rgb-std/commit/251cecf908f61a92df534879da1c693cd3c12cb9)

I will test now.

PR #148, opened 2024-02-15
New transfer id workflows

dr-orlovsky:


crisdut (commented):
Now, the following error is occurring:

```
The transfer the actual data doesn't match the provided id.\nActual id: transfer:Cs7AAr-dHR89bcS-nJ86BwiN-QXHcMKXj-yRvgKsb3-1D8LnV#baker-manual-patent.\nExpected id: transfer:ARZGSt-jNh7wRYp-zK9cyLXX-o2psW26n-GWf9qedh-j3JxY1#clever-dispute-zipper. presents inconsistency
```

Just a testing:  If we change `self.id` in the [line](https://github.com/RGB-WG/rgb-std/pull/148/files#diff-2a54dc4f21030f78ab797cc76156cb322d2d8ecede7e48122114a6ca18d1495aR257) to `self.bindle_id()`, the error is corrected.

stl/RGB20@0.1.0.sta
```diff
@@ -1,11 +1,11 @@
 -----BEGIN STRICT TYPE LIB-----
-Id: urn:ubideco:stl:GVz4mvYE94aQ9q2HPtV9VuoppcDdduP54BMKffF7YoFH
+Id: GVz4mvYE94aQ9q2HPtV9VuoppcDdduP54BMKffF7YoFH#prince-scarlet-ringo
```
cryptoquick, line 2: Were we supposed to lose the URN prefix?

stl/RGB20@0.1.0.sty
```diff
@@ -4,16 +4,18 @@
...
 
+@context
 typelib RGB20
 
 -- no dependencies
 
--- urn:ubideco:semid:BsYakBZMzKBDfgpfhXt1p81Q1WT27V8U7LvHG4NPmGmM#montana-torso-pierre
-data Error            :: supplyMismatch:1 | nonEqualAmounts:2 | invalidProof:3 | insufficientReserves:4
-                       | insufficientCoverage:5 | issueExceedsAllowance:6
+@mnemonic(montana-torso-pierre)
```
cryptoquick, line 16: What's the reasoning behind providing less data here?

cryptoquick (commented):
Some questions

src/containers/transfer.rs
```diff
@@ -66,39 +73,34 @@ impl TransferId {
...
+
+    fn commit_encode(&self, e: &mut CommitEngine) {
+        e.commit_to(&self.version);
+        e.commit_to(&self.transfer);
+
+        e.commit_to(&self.contract_id());
+        e.commit_to(&self.genesis.disclose_hash());
+        e.commit_to(&TinyOrdSet::from_iter_unsafe(
+            self.ifaces.values().map(|pair| pair.iimpl.impl_id()),
+        ));
```
crisdut, line 86: All items must be named types.

crisdut (changes requested):

crisdut (changes requested):
We have good news: all tests ran successfully in my environment, so I believe that everything is now working.

However, the only bug I found occurs when we try converting a `Bindle<Transfer>` to `string` and `string` to `Bindle<Transfer>`:

```
The transfer the actual data doesn't match the provided id.
Actual id: transfer:CeJAm5-nw1WxMzV-81DtaJSa-CTgjJ1MS-g9NmSjvi-6vMkT3#bonanza-shelter-organic.
Expected id: transfer:CgtX9T-9uVVt1Rq-3dPAs62P-CegfHe1Y-Gjs8D8Uu-onnw1L#rebel-justice-lithium. presents inconsistency
```

I think we need make some changes [here](https://github.com/RGB-WG/rgb-std/blob/5817be7600c38fcfc5e4499ca165dbfe3444aea1/src/containers/bindle.rs#L257), from `writeln!(f, "Id: {:-#}", self.id)?;` to `writeln!(f, "Id: {:-#}", self.bindle_id())?;`

If this change, this [validation](https://github.com/RGB-WG/rgb-std/blob/5817be7600c38fcfc5e4499ca165dbfe3444aea1/src/containers/bindle.rs#L237C16-L237C25) works.

dr-orlovsky:
@crisdut thank you for figuring out the source of the problem! You are right: `bindle.id` is the ID for the bindle _content_, while `bindle.bindle_id()` commits to both content AND signatures...

Shit, these id's are really getting out of control...

dr-orlovsky:
No, I  am wrong: id and bindle_id should be the same thing! At least according to the code: https://github.com/RGB-WG/rgb-std/blob/5817be7600c38fcfc5e4499ca165dbfe3444aea1/src/containers/bindle.rs#L171

Any guesses why we have this bug? I can easily fix with your suggestion above, but I am afraid we are missing some deep bug here...

crisdut:
> Any guesses why we have this bug? I can easily fix with your suggestion above, but I am afraid we are missing some deep bug here...

I found the reason.


When we create the bindle<transfer>, we use the Bindle::new method:

```rust
pub struct Bindle<C: BindleContent> {
    id: C::Id,
    data: C,
    sigs: TinyVec<Cert>,
}

impl<C: BindleContent> Bindle<C> {
    pub fn new(data: C) -> Self {
        Bindle {
            id: data.bindle_id(),
            data,
            sigs: empty!(),
        }
    }
}
```

In some cases, we modify the tranfer, and this changes producing another id when we execute bindle_id and transfer_id, but Bindle::Id remain fixed.

Example:

https://github.com/RGB-WG/rgb/blob/master/src/pay.rs#L337-L353

My recommendation is remove the directly access by id, and use always bindle_id.

What's you think?

dr-orlovsky:
@crisdut I did this way in https://github.com/RGB-WG/rgb-std/pull/148/commits/0a9c92ef073ac3c3c61390cebe94aa0da341e86d and updated RGB wallet in master. Please let me know if everything works now - and I will merge this PR

PR #156, opened 2024-03-10
Assignment interfaces for rgb20

cymqqqq:
new assignment interfaces for rgb20

### DIFF
--- a/src/interface/contract.rs
+++ b/src/interface/contract.rs
@@ -82,10 +82,6 @@ impl KnownState for AllocatedState {}
 pub type OwnedAllocation = OutputAssignment<AllocatedState>;
 pub type RightsAllocation = OutputAssignment<VoidState>;
 pub type FungibleAllocation = OutputAssignment<Amount>;
-pub type InflationAllowanceAllocation = OutputAssignment<Amount>;
-pub type UpdateRights = OutputAssignment<VoidState>;
-pub type BurnEpoch = OutputAssignment<VoidState>;
-pub type BurnRight = OutputAssignment<VoidState>;
 pub type DataAllocation = OutputAssignment<DataState>;
 pub type AttachAllocation = OutputAssignment<AttachedState>;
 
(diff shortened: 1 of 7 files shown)


src/interface/contract.rs
```diff
@@ -84,6 +82,10 @@ impl KnownState for AllocatedState {}
 pub type OwnedAllocation = OutputAssignment<AllocatedState>;
 pub type RightsAllocation = OutputAssignment<VoidState>;
 pub type FungibleAllocation = OutputAssignment<Amount>;
+pub type InflationAllowanceAllocation = OutputAssignment<Amount>;
+pub type UpdateRights = OutputAssignment<VoidState>;
+pub type BurnEpoch = OutputAssignment<VoidState>;
+pub type BurnRight = OutputAssignment<VoidState>;
```
dr-orlovsky, line 88: I think we do not need these newtypes (and certainly not in the global namespace, since they are RGB20-specific).

Just use `FungibleAllocation`, and `RightsAllocation` instead in the code you have added. The rest looks good

dr-orlovsky (changes requested):
LGTM except one commend (please check it).

Also you need to run `fmt` and `clippy`, as in other PRs, and make PRs in a way that they do not share the same history

src/interface/contract.rs
dr-orlovsky: The other PR is merged, so can you pls update this PR with removing these type names - and I will merge it and proceed to the next ones

dr-orlovsky: @cymqqqq your PRs share the same commits, thus if you update one of them the other will have a diverging history. I suggest to split the commits into PRs (such that each one has one or two) - or just give the one PR with all commits. Otherwise the process will be quite complex

cymqqqq: I have updated this PR, and thanks, I will try my best to just give one PR later. :)

dr-orlovsky (approved):

Issue #151, opened 2024-01-22
Beneficiary vs assetOwner field name for transitions


Matrix-Zhang:
```
Error: field name beneficiary is unknown to the contract interface
```

dr-orlovsky:
This should be fixed when https://github.com/RGB-WG/rgb-std/pull/162 is merged

PR #162, opened 2024-03-12
Interface refactoring

dr-orlovsky:
- [x] Remove state aliases from operations
- [x] Increase iface version
- [x] Use lowercase names for operations
- [x] Refactor errors: take error variant id from pre-defined register
- [x] Move timestamp into genesis
- [x] Add issuer identity and certificate to genesis
- [x] Refactor asset terms (ricardian contract, attachment)
- [x] Refactor asset name (flatten it)
- [x] Question the need in RGB25 (it is needed)
- [x] Add explicit media files to the mediakit in supplement
- [x] Add signature for the supplement
- [x] ~~RGB20 has isufficientReserves but doesn't provide reserves state~~ (it does, in metadata)
- [x] Create iface compiler verifying iface consistency
- [x] Export interfaces into Contractum code
- [x] Add support for interface inheritance


dr-orlovsky:
@zoedberg can I merge this? It is required for https://github.com/RGB-WG/rgb-std/issues/151

I mean I understand that it's hard to review; I rather expect you to check the checklist and ask if you see any red flags there. If not - I will merge and we will debug over time all things which may be wrong here

zoedberg (changes requested):
This doesn't compile in rgb-lib because it breaks compilation of rgb-schemata. I don't think we should merge this without first having a PR for rgb-schemata.

dr-orlovsky:
@zoedberg right, fully forgot about that. Here it is: https://github.com/RGB-WG/rgb-schemata/pull/36

I will also work on the RGB wallet changes

src/stl/specs.rs
```diff
@@ -498,21 +421,21 @@ impl Attachment {
...
-impl StrictSerialize for ContractData {}
-impl StrictDeserialize for ContractData {}
+impl StrictSerialize for AssetTerms {}
+impl StrictDeserialize for AssetTerms {}
 
-impl ContractData {
+impl AssetTerms {
     pub fn from_strict_val_unchecked(value: &StrictVal) -> Self {
         let terms = RicardianContract::from_str(&value.unwrap_struct("terms").unwrap_string())
             .expect("invalid terms");
```
zoedberg, line 434:
```suggestion
        let text = RicardianContract::from_str(&value.unwrap_struct("text").unwrap_string())
            .expect("invalid text");
```

src/stl/specs.rs
```diff
@@ -498,21 +421,21 @@ impl Attachment {
...
         let media = value
             .unwrap_struct("media")
             .unwrap_option()
             .map(Attachment::from_strict_val_unchecked);
-        Self { terms, media }
+        Self { text: terms, media }
```
zoedberg, line 439:
```suggestion
        Self { text, media }
```

zoedberg (changes requested):
I've updated rgb-lib to try the PRs refactoring the interfaces (https://github.com/RGB-WG/rgb-std/pull/162, https://github.com/RGB-WG/rgb-schemata/pull/36 and https://github.com/RGB-WG/rgb/pull/154) but when doing:
```rust
let iface_nia = Rgb20::from(contract_iface);
let terms = iface_nia.contract_terms();
```
I encounter the following panic:
```rust
thread 'wallet::test::send::success' panicked at /home/zoe/.cargo/registry/src/index.crates.io-6f17d22bba15001f/strict_types-2.7.0-beta.2/src/value/val.rs:399:32:
StrictVal struct doesn't have field named terms
```

I've found a fix for this, see the provided suggestions.

After fixing this I've found another issue: before I was retrieving the asset timestamp with `contract.global("created")`, but I see this PR removed the `created` global state. Why this has been removed and how can we now retrieve this information?

dr-orlovsky:
Thanks for finding a fix.

Regarding the created timestamp, it is moved into the genesis itself, as was proposed in [RCP-240227A](https://github.com/RGB-WG/RFC/issues/1) and discussed in Telegram.

If you need to have an access to it, you can just read `timestamp` property on genesis: https://github.com/RGB-WG/rgb-core/blob/e04cfffe989642eb03db4d065c6f53c1d6d218ba/src/contract/operations.rs#L317

zoedberg (approved):
Ok, thanks, sorry I missed that. I've updated rgb-lib timestamp extraction code and now all tests are passing. ACK 00f19831580384c2a1060330d0e21711fbbc3a17

PR #166, opened 2024-03-15
RCP-240313B: Abstract script and type libraries

dr-orlovsky:
This is built on top of https://github.com/RGB-WG/rgb-core/pull/221

The approach is:
- keep script and type libraries in stash as separate objects
- create a concept of generic container: data exchanged between RGB users, which contain required type and script libraries in addition to the main data
- let these containers contain also detached signatures on type and script libs (and other content)
- introduce concept of a Kit - a package of schema, together with interfaces, implementations and libraries, needed to issue a contract


dr-orlovsky:
Closed in favor of https://github.com/RGB-WG/rgb-std/pull/187

PR #168, opened 2024-03-24
Do not update blindings for blank transitions

### LINKED ISSUE #149: CFA transfer (with a NIA blank) fails
Using this [branch](https://github.com/nicbus/rgb-sandbox/tree/cfa) of rgb-sandbox, which includes support for CFA assets, I fail to transfer a CFA asset. The UTXO used as input for the transfer also has another (NIA) asset allocated, so a blank transition is also created. I get this error:
```
Validation failures:
- operation 1cc8530db258f5be43bbbb2afbf6787874c286267154d26d1c6b0d0a2e30a591 is invalid: unspecified error
```

Said branch of rgb-sandbox uses rgb-std from branch `v0.11` with the added fix for the blank contract ID (originally https://github.com/RGB-WG/rgb-std/pull/140).

Experimenting with the sandbox and rgb-lib, I found out that transfers succeed if I disable the `BlindingFactor` update that happens in `rgb-std`'s `compose_deterministic()`.
This can be tested using this [branch](https://github.com/nicbus/rgb-sandbox/tree/cfa_no_blank_blinding_update) of rgb-sandbox.

dr-orlovsky:
Should close #149

### DIFF
--- a/src/persistence/inventory.rs
+++ b/src/persistence/inventory.rs
@@ -901,12 +901,9 @@ pub trait Inventory: Deref<Target = Self::Stash> {
         for (id, opouts) in spent_state {
             let mut blank_builder = self.blank_builder(id, iface.clone())?;
             let mut outputs = Vec::with_capacity(opouts.len());
-            for ((opout, output), mut state) in opouts {
+            for ((opout, output), state) in opouts {
                 let seal = output_for_assignment(id, opout.ty)?;
                 outputs.push(output);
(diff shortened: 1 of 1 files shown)


dr-orlovsky:
Motivation: generating new blinding factor for Pedersen commitment is meaningless, since if you have just one input and one output the output blinding factor MUST be equal to the one used in input - what is done anyway in `complete_transition` method

PR #172, opened 2024-03-25
Interface inheritance

dr-orlovsky:


yanganto (changes requested):
Some `continue` arms can be removed and use `break` instead.  Others look good to me.

yanganto (approved):

PR #177, opened 2024-04-07
ci: bump msrv

yanganto:
`rgb-core v0.11.0-beta.5` requires rustc 1.70.0 or newer, so the msrv bump.  The msrv mentioned in README is incorrect with the 1.66 version, but the ci is using 1.69.0 as msrv.

To avoid this issue in the future, the msrv will be based on `rust-version` of `Cargo.toml` as the single source of truth, the readme will refer to the file, and the ci will read the file and check on the build of msrv.


yanganto:
covered by #181

PR #178, opened 2024-04-07
v0.11 Beta 6 vs Beta5: sum of all changes

dr-orlovsky:


src/containers/util.rs
```diff
@@ -49,20 +48,12 @@ pub struct TerminalDisclose {
...
 pub struct Terminal {
     pub seals: SmallOrdSet<XChain<TerminalSeal>>,
-    pub witness_tx: Option<XPubWitness>,
 }
 
 impl Terminal {
     pub fn new(seal: XChain<TerminalSeal>) -> Self {
         Terminal {
             seals: small_bset![seal],
-            witness_tx: None,
```
cymqqqq, line 59: Why do we remove the witness_tx here?

dr-orlovsky: Since now it is present directly in the `SealWitness` structure for all bundles, not just terminal ones.

PR #182, opened 2024-04-09
Abstract implementations to more generic interfaces using inheritance

dr-orlovsky:


yanganto (changes requested):
It seems that `filter_map` is a better way to handle this.

yanganto:
> Can you elaborate? As I see it just increases number of code lines
I did PR #188, and the code line was reduced.

yanganto (approved):

PR #185, opened 2024-04-10
iface: break if overflow

yanganto:
ref #172


dr-orlovsky (commented):
Seems we can go further: remove overflow variable and move if check to be `if self....is_err(){ break }`

dr-orlovsky (approved):
ACK 8b13a7a231913914362dccfe4f97971a6b7c4e52

Issue #111, opened 2023-11-22
Provide an API for deterministic contract generation


oneforalone:
I tried to make every arguments to const value, however, each time i compiled it, the contract id changes. And after take tracing the code, found the contract id is generate from lib `commit_verify`'s function `commitment_id`. The function shows that the id is generate from a sha256 of a 32 bytes variable `TAG`, while i found that `TAG`s variables are defined as const, so, why the contract id would become different with the same arguments and the same codes? Or is there an way to generate a deterministic contract id with the same codes/arguments?

dr-orlovsky:
No, it is not possible. Each contract is unique - at least as long as it contains a timestamp (which includes milliseconds).

oneforalone:
I figure it out that RGB20 contract id is based on `fungible` field which is a `Confined<BTreeMap<u16, Confined<HashMap<BuilderSeal<Seal>, RevealedValue, RandomState>, 1, U8>, Global>, ZERO, U8>` type, and there's `RandomState` in it, thus, there's no way to make generate a deterministic contract id. And if the `RandomState` is removed or deterministic, then the blind seal would be meaningless, am I right?

dr-orlovsky:
No, `RandomState` is the Rust thing, which has nothing to do with RGB or indeterminism. It is used by the `BTreeMap` as an allocator.

However, inside `RevealedValue` there is an automatically generated blinding factor, and you are right, that makes the contract id different each time. And there is no way to remove it. However, this is not the only reason: even if it is removed, you will have different contract ids due to different timestamps.

Why do you need a deterministic contract generation?

oneforalone:
> Why do you need a deterministic contract generation?

Cause in my point of view, contract id should be the same if and only if the contract contents is the same. I mean, it's just a hash of the contract content.

Actually, i tried to set Timestamp to a constant, and for the `RevealedValue`, yeah, there's a blind factor which is a `secp256k1_zkp::SecretKey`, and i do set this argument to a constant with a fixed timestamp. The contract id still changes. Am i missing something?

dr-orlovsky:
If you are using current master/beta v0.11 there is also a thing called `AssetTag` which is there for Liquid compatibility.

Anyway, can you share your code so I can check how exactly you are doing that?

oneforalone:
> If you are using current master/beta v0.11 there is also a thing called `AssetTag` which is there for Liquid compatibility.

I'm still on version 0.10.2, not using `AssetTag` attribute.

[...]

PR #190, opened 2024-04-16
add add_global_state_det

zoedberg:
This PR adds the possibility to add global state in a determistic way (necessary for the deterministic contract ID creation, see https://github.com/RGB-WG/rgb-std/issues/111)

### DIFF
--- a/src/interface/builder.rs
+++ b/src/interface/builder.rs
@@ -25,6 +25,7 @@ use std::collections::{BTreeMap, HashSet};
 
 use amplify::confinement::{Confined, SmallOrdSet, TinyOrdMap, U16};
 use amplify::{confinement, Wrapper};
+use bp::secp256k1::rand::random;
 use chrono::Utc;
 use invoice::{Allocation, Amount};
 use rgb::{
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
Thank you! I think we also need to assert that we do mot add deterministic state in non-deterministic builders - and then other way around. You can check how we do that in other method checking `deterministic` field

zoedberg:
> Thank you! I think we also need to assert that we do mot add deterministic state in non-deterministic builders - and then other way around. You can check how we do that in other method checking `deterministic` field

I'm not sure I understood what you exactly mean but I gave it a try anyway. I've also updated rgb-schemata PR accordingly

dr-orlovsky:
Ok, sorry, I was wrong two times:

First, this problem is a bug in RGB Core and must be fixed there: the global state must not have a salt or concealment. Are is the fix: https://github.com/RGB-WG/rgb-core/pull/241

Second, deterministic flag was present in RGB20 helpers, not this builders. I recognized it has to be added at the level of `OperationBuilder`, which I did in https://github.com/RGB-WG/rgb-std/pull/191

Closing this PR since we have to use those different approaches. Sorry for not understanding it in the original review

PR #191, opened 2024-04-17
Complete deterministic builder

dr-orlovsky:
This is an alternative to https://github.com/RGB-WG/rgb-std/pull/190#issuecomment-2060902799

### DIFF
--- a/src/interface/builder.rs
+++ b/src/interface/builder.rs
@@ -153,6 +153,21 @@ impl ContractBuilder {
         }
     }
 
+    pub fn deterministic(
+        iface: Iface,
+        schema: Schema,
+        iimpl: IfaceImpl,
+        types: TypeSystem,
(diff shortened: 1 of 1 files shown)


zoedberg (approved):
Tested, it works so I'm approving this.
Just a small note: trying to use deterministic methods on a builder constructed with `ContractBuilder::with` returns the error ```to add owned state in deterministic way the builder has to be created using `*_det` constructor```, which IMO makes the user think he should call `ContractBuilder::with_det`, but that doesn't exist and the correct way of constructing a deterministic builder is `ContractBuilder::deterministic`. So maybe we should rename the `deterministic` method to `with_det` or change the error messages of deterministic methods.

dr-orlovsky:
Sure! Done in 0a392e705e1cfd246508021467f2963552d366d7

PR #202, opened 2024-05-20
fix ProofOfReserves::from_strict_val_unchecked

zoedberg:
Call to `contract.token_data()` causes the following error:
```
StrictVal expected to be a string but holds non-string value `(txid=(0xcb6c173a05cc4f9432a26b2048b0021e2e9d044980f406d61bf39d3077e5a3e5), vout=(0))`
```

This PR fixes `ProofOfReserves::from_strict_val_unchecked`, which is the method causing the error.

### DIFF
--- a/src/stl/chain.rs
+++ b/src/stl/chain.rs
@@ -19,10 +19,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::str::FromStr;
-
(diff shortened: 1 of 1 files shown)


zoedberg:
@dr-orlovsky I think this should be merged before releasing beta 6, could you please have a look at it?

Issue #119, opened 2023-12-30
Chunk addresses inside invocies


dr-orlovsky:

dr-orlovsky:
Closed in #203

PR #203, opened 2024-05-24
Support custom close methods in vout-based invoices

### LINKED ISSUE #198: witness beneficiary not embedding close method
During Viareggio meeting we reported that the `WitnessVout` beneficiary doesn't embed the desired close method. It was said that this is a bug because a receiver should be able to decide this. Currently it's up to the sender to decide which close method to use in that case.

@dr-orlovsky I think we should fix this before 0.11 preview release

dr-orlovsky:
Closes #198 and #119

I decided to address this two issues simultaneously, since it is impossible to reliable parse chunked variable-length strings; thus, in order to get chunking I had to switch to Base64 and fixed-length strings. Since we had to move from bitcoin address format with that, I used the approach as a base to add closing method in a compact form (one byte).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,8 +142,7 @@ checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 [[package]]
 name = "baid64"
 version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd5e666513565d0a35116973541fbcb35700619316e1865cb27c7dba5de98668"
+source = "git+https://github.com/UBIDECO/rust-baid64#23b3dd47058cd5d1a338b6dbf48cefb5d1441d59"
 dependencies = [
  "amplify",
  "base64",
@@ -173,10 +172,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
-name = "bitcoin-internals"
-version = "0.2.0"
+name = "bitcoin-io"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9425c3bf7089c983facbae04de54513cce73b41c7f9ff8c845b54e7bc64ebbfb"
+checksum = "340e09e8399c7bd8912f495af6aa58bea0c9214773417ffaa8f6460f93aaee56"
 
 [[package]]
 name = "bitcoin-private"
@@ -186,11 +185,11 @@ checksum = "73290177011694f38ec25e165d0387ab7ea749a4b81cd4c80dae5988229f7a57"
 
 [[package]]
 name = "bitcoin_hashes"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1930a4dabfebb8d7d9992db18ebe3ae2876f0a305fab206fd168df931ede293b"
+checksum = "bb18c03d0db0247e147a21a6faafd5a7eb851c743db062de72018b6b7e8e4d16"
 dependencies = [
- "bitcoin-internals",
+ "bitcoin-io",
  "hex-conservative",
 ]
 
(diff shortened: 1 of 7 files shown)


zoedberg (commented):
@dr-orlovsky To try this also rgb-runtime needs to be updated:
```
   Compiling rgb-runtime v0.11.0-beta.5 (/mnt/dmc/zoe/work/bitfinex/rgb-integration-tests/rgb)
error[E0308]: mismatched types
   --> rgb/src/pay.rs:208:34
    |
208 |                     Address::new(payload, invoice.address_network()),
    |                     ------------ ^^^^^^^ expected `AddressPayload`, found `Pay2Vout`
    |                     |
    |                     arguments to this function are incorrect
    |
note: associated function defined here
   --> /mnt/dmc/zoe/work/bitfinex/rgb-integration-tests/bp-std/invoice/src/address.rs:105:12
    |
105 |     pub fn new(payload: AddressPayload, network: AddressNetwork) -> Self {
    |            ^^^

error[E0599]: no method named `script_pubkey` found for struct `Pay2Vout` in the current scope
   --> rgb/src/pay.rs:224:27
    |
224 |                 Some(addr.script_pubkey())
    |                           ^^^^^^^^^^^^^ method not found in `Pay2Vout`
    |
help: one of the expressions' fields has a method of the same name
    |
224 |                 Some(addr.address.script_pubkey())
    |                           ++++++++

error[E0599]: no method named `script_pubkey` found for struct `Pay2Vout` in the current scope
   --> rgb/src/pay.rs:250:30
    |
250 |                 let s = addr.script_pubkey();
    |                              ^^^^^^^^^^^^^ method not found in `Pay2Vout`
    |
help: one of the expressions' fields has a method of the same name
    |
250 |                 let s = addr.address.script_pubkey();
    |                              ++++++++

error[E0599]: no method named `script_pubkey` found for struct `Pay2Vout` in the current scope
   --> rgb/src/pay.rs:301:30
    |
301 |                 let s = addr.script_pubkey();
    |                              ^^^^^^^^^^^^^ method not found in `Pay2Vout`
    |
help: one of the expressions' fields has a method of the same name
    |
301 |                 let s = addr.address.script_pubkey();
    |                              ++++++++

Some errors have detailed explanations: E0308, E0599.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `rgb-runtime` (lib) due to 4 previous errors
```

dr-orlovsky:
I am updating it, but in the meanwhile can you say are you on with the approach?

zoedberg:
> I am updating it, but in the meanwhile can you say are you on with the approach?

I agree with the approach

dr-orlovsky:
@zoedberg: here you are: https://github.com/RGB-WG/rgb/pull/202

zoedberg:
I've tried using https://github.com/RGB-WG/rgb-std/pull/203 and https://github.com/RGB-WG/rgb/pull/202, now it compiles but the call to the `pay` method gives an error when trying to pay an invoice with a witness beneficiary.

The error appears only when a TapretFirst wallet tries to pay an invoice containing a witness benecifiary with OpretFirst as close method and when an opret wallet tries to pay an invoice containing a witness beneficiary with TapretFirst as close method.

Here the error of the first described case:
```
called `Result::unwrap()` on an `Err` value: Completion(Stock("outpoint bc:tapret1st:332e7bb119f871b0128cad67f2e698504669ffd95e78bf1022d256acd9ab6b57:1 is not part of the contract rgb:p7vHgbEH-mFdX8Sp-A!PIzpI-nXaEhyM-PAA6srV-PTUYqA8.\nIt may happen due to RGB standard library bug, or indicate internal stash inconsistency and compromised stash data storage."))
```
and here the error of the second described case:
```
called `Result::unwrap()` on an `Err` value: Completion(Stock("outpoint bc:opret1st:ad1afc53b6331104b75bddcb6f791231d2d60b883a94da525be209e291909b0e:0 is not part of the contract rgb:SKcxpcA2-PPfEJd!-9oFkksd-sG7P7rx-CLhcVbt-33zg8oI.\nIt may happen due to RGB standard library bug, or indicate internal stash inconsistency and compromised stash data storage."))
```

dr-orlovsky:
Can you please give me a call stack when the error does happen? I am unable to understand the reason of it so far...

dr-orlovsky:
The issue was fixed with https://github.com/RGB-WG/rgb/pull/202/commits/7717d2466f90646a53ada4799fc67a65267a9e28

PR #204, opened 2024-05-25
Coin amount type

### LINKED ISSUE #120: Use readable precision in invoice amount
The invoice for 100.00 of an asset with precision 2 or more should look like `rgb:/RGB20/100.00+tb:utxob:...` (independently from whether the precision is 2 or 8) and not like `rgb:/RGB20/10000+tb:...` (if the precision is 2) or like `rgb:/RGB20/10000000000+tb:...` (if the precision is 8).

This should help people to visualize how much they are actually paying.

dr-orlovsky:
Coin amount is a helper type for displaying values with precision

### DIFF
--- a/invoice/src/amount.rs
+++ b/invoice/src/amount.rs
@@ -23,12 +23,14 @@ use std::cmp::Ordering;
 use std::fmt;
 use std::fmt::{Display, Formatter, Write};
 use std::iter::Sum;
+use std::num::{ParseIntError, TryFromIntError};
+use std::str::FromStr;
 
 use bp::Sats;
 use rgb::{FungibleState, KnownState, RevealedValue};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-use strict_encoding::{StrictDeserialize, StrictSerialize};
+use strict_encoding::{StrictDeserialize, StrictSerialize, VariantError};
 use strict_types::StrictVal;
 
 use crate::LIB_NAME_RGB_CONTRACT;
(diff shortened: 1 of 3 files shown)


nicbus (commented):
LGTM in general.

I'm not sure about the usage of the ` `, `_`, `` ` `` and `'` characters as fillers, though, as they look a bit strange to me in this context.
I'd be ok with just `_` (or no fill) but I don't understand why we need more than one and why those characters specifically.

dr-orlovsky:
@nicbus `CoinAmount` is intended to be used not just in invoices, but in different parts of the UI. `_` in UI look really bad - and existing typography across the globe uses spaces, ticks or reverse ticks instead - this is kind of standard. Thus, they are supported.

nicbus:
> CoinAmount is intended to be used not just in invoices, but in different parts of the UI. _ in UI look really bad

I agree that `_` is not optimal in some UIs and am fine with alternatives. I just find the other characters a little awkward.

That being said, I guess most UIs will handle grouping internally (possibly locale-based) and default to no separator when interfacing with libraries.

> existing typography across the globe uses spaces, ticks or reverse ticks instead - this is kind of standard. Thus, they are supported.

I couldn't find info about usage of the `` ` `` character as a standard.

I found some info about the usage of the ` ` char in [ISOSTS](https://www.iso.org/schema/isosts/v1.1/doc/index.html) (under `Attributes` > `gsep`).

[Wikipedia](https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping) lists the typical (but deprecated) locale-dependent group separators as `,` (when `.` is used as decimal separator) and `.` (when `,` is used as decimal separator), which I don't like because they can be confusing (and I guess this was your reasoning as well). Other than that, the `'` is mentioned once (but it don't seem widely used, I might be wrong) and the space (thin or regular) character is well described.

From these sources, I get that the preferred grouping separator should be the "thin space", with the "regular space" and the underscore as the alternatives.

Can you provide some info on why `'` and especially `` ` `` are used as standard group separators?

dr-orlovsky:
> Can you provide some info on why ' and especially ` are used as standard group separators?

With this PR there is nothing "standard separator": the `Display` implementation uses any fill character which is provided as a part of formatting string. It is only `FromStr` and tests which specifically accounts for `'` and backtick characters - but that doesn't make them "standard"

I think for now it will be better to open an issue with an option which you suggest as the best fix if you believe this is not a good practice.

nicbus:
Based on mentioned Wikipedia info and the lack of evidence supporting the use of other characters as group separators, I would simply drop every character in the set except for the underscore and the space characters.

The tick and backtick can give rise to issues in a CLI, as they're special characters in shells, so it makes sense to avoid them.

@dr-orlovsky I can open a PR with these changes if you agree

dr-orlovsky:
I agree

Issue #208, opened 2024-06-08
Possible misuse of method `fluent_uri::Path::segments`


yescallop:
Hi, the maintainers of `rgb-std`. I'm the author of `fluent-uri`. I notice that this repo contains code calling the method `fluent_uri::Path::segments`: https://github.com/RGB-WG/rgb-std/blob/9e23bd272a989c5f0315f458565abfe3c1f7b2a0/invoice/src/parse.rs#L333-L337

Which might not work as intended. It currently accepts both absolute and rootless paths, for example `/foo/bar` and `foo/bar`.

Whether this behavior is intended or not, I think I should expand the method's documentation so that it's less likely to be misused. I have changed it to the following: https://github.com/yescallop/fluent-uri-rs/blob/db4cde70754126cdc75425fefd655d1c88904a82/src/encoding/mod.rs#L382-L414

Thanks for depending on my crate `fluent-uri` and I'll continue to make sure it offers a good user experience.

As an aside, I find that the following code looks a bit strange on line 274: https://github.com/RGB-WG/rgb-std/blob/9e23bd272a989c5f0315f458565abfe3c1f7b2a0/invoice/src/parse.rs#L268-L277

It might not work as intended as I imagine the `rgb:` prefix should be written in both branches, or I could be wrong since I don't know much about the project. Still, I bring this up just in case.

dr-orlovsky:
Dear @yescallop! Thank you very much for taking time reviewing our code and writing such a detailed fix proposal!

dr-orlovsky:
BTW you are right about the needed check on whether the path is absolute. I did a PR here: https://github.com/RGB-WG/rgb-std/pull/209

As for the `rgb:` prefix in the second suggestion, everything is correct since in the first branch it is printed as a part of `ContractId` `Display` implementation

PR #209, opened 2024-06-08
Require invoice URIs not to contain authority

dr-orlovsky:
Addresses review https://github.com/RGB-WG/rgb-std/issues/208

### DIFF
--- a/invoice/src/parse.rs
+++ b/invoice/src/parse.rs
@@ -75,6 +75,10 @@ pub enum InvoiceParseError {
     /// invalid invoice.
     Invalid,
 
+    /// RGB invoice must not contain any URI authority data, including empty
+    /// one.
+    Authority,
+
     /// contract id is missed from the invoice.
     ContractMissed,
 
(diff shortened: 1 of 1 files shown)


dr-orlovsky:
@nicbus I am asking for a post-merge review just in case - I am pretty sure everything is correct and have to release beta-6 before traveling to BTC Prague

yescallop:
Thanks for your attention to my report. I notice that the code in this fix accesses the `auth` field of `fluent_uri::internal::Data` which is not part of `fluent-uri`'s public API. It was my fault to make it `pub`. Would you switch to using `Uri::authority` instead so that it doesn't break things for me to release a new version that set those fields as private later?

dr-orlovsky:
@yescallop thank you for letting me know. Fixed in https://github.com/RGB-WG/rgb-std/commit/0046fc446c6150b16f8df3bba9b048d3de267961

Also I think method like `has_authority()` would be handy

PR #210, opened 2024-06-12
fix bug: Pay2Vout's to_baid64_payload and try_from

eauxxs:


eauxxs:
#### example:
```
wpkh script: 00148833bd1dc92076f2ad60c4061c57667d8a62657d
// origin
&origin = Pay2Vout {
    method: OpretFirst,
    address: Wpkh(
        WPubkeyHash(
            Array<20>(8833bd1dc92076f2ad60c4061c57667d8a62657d),
        ),
    ),
}
// tmp = Pay2Vout::from(&origin.to_string())
&tmp = Pay2Vout {
    method: OpretFirst,
    address: Wpkh(
        WPubkeyHash(
            Array<20>(00148833bd1dc92076f2ad60c4061c57667d8a62),
        ),
    ),
```

zoedberg (approved):
I confirm that with beta6 there's a bug with witness transfers, here's the error message:
```
thread 'wallet::test::send::witness_success' panicked at /home/zoe/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rgb-invoice-0.11.0-beta.6/src/parse.rs:258:14:
address payload always less than 32 bytes: Error { kind: WriteZero, message: "failed to write whole buffer" }
```
I also confirm that this PR fixes the issue, so tACK a0da52ad4da164a08082cc75c0e6812a5458a0be.

eauxxs:
@zoedberg thanks for your review. By the way, I have adapted the latest rgb beta6 for rgb-lib, and it has passed all tests after this pr. So if you need it, I can try to raise a PR, and hope to reduce your work.

zoedberg:
> @zoedberg thanks for your review. By the way, I have adapted the latest rgb beta6 for rgb-lib, and it has passed all tests after this pr. So if you need it, I can try to raise a PR, and hope to reduce your work.

Thanks for the offer but I've already updated rgb-lib to the beta 6 (as told you here https://github.com/RGB-Tools/rgb-lib/issues/48). Anyway, we'll not publish that commit since the tests with witness transfers are not working. We'll wait for beta 7

dr-orlovsky (changes requested):
Thank you for finding and fixing the issue. Unfortunately we need one more thing: reduce the length of the array from 35 to 34 bytes everywhere. Can you do that or you prefer me adding commit to your PR?

dr-orlovsky (commented):
Can you please explain why we can't continue using `self.address.script_pubkey()` and instead need to do the conversion inside the `match` block?

eauxxs:
@dr-orlovsky Because address.script_pubkey() does not return only the payload as you think, ScriptPubkey::as_slice contains prefixes, such as OP_0 and OP_PUSHBYTE_20.

https://github.com/BP-WG/bp-std/blob/8177aab7c0badf5295b6b9664d2f652a94f7093e/invoice/src/address.rs#L346-L352

So we should use the match method to extract the payload inside. Because of the special nature of the tr address, its processing should apply for a 32-byte memory in the stack to store the payload, and cannot directly as_ref<[u8]>.

dr-orlovsky:
@eauxxs can you please rebase this on top of the master which now contains the length and offset fixes - to see what specifically we need to fix in order to get P2TR working

eauxxs:
@dr-orlovsky Sorry, I am very confused. This fix for pay2vout conflicts with #214. I think #214 is wrong, I don't know why you would merge that PR after I explained the main error to you.

here is test code, you can run it:
```
#[test]
fn pay2vout_parse() {
    use crate::AddressPayload;
    let p = Pay2Vout {
        method: bp::dbc::Method::OpretFirst,
        address: AddressPayload::Pkh([0xff; 20].into()),
    };
    assert_eq!(Pay2Vout::from_str(&p.to_string()).unwrap(), p);
}
```

```
---- parse::test::pay2vout_parse stdout ----
thread 'parse::test::pay2vout_parse' panicked at invoice/src/parse.rs:757:9:
assertion `left == right` failed
  left: Pay2Vout { method: OpretFirst, address: Pkh(PubkeyHash(Array<20>(14ffffffffffffffffffffffffffffffffffffff))) }
 right: Pay2Vout { method: OpretFirst, address: Pkh(PubkeyHash(Array<20>(ffffffffffffffffffffffffffffffffffffffff))) }
```
https://github.com/RGB-WG/rgb-std/blob/6874b25273c5761275ddce37428e6e9b5972e960/invoice/src/parse.rs#L256-L257

If you must use self.address.script_pubkey().as_slice(), for p2pkh addresses, you should skip the first 3 bytes, because the prefix of p2pkh is OP_DUP OP_Hash160 OP_PUSHBYTES_20.

invoice/src/invoice.rs
```diff
@@ -212,17 +212,17 @@ impl Pay2Vout {
...
-    fn try_from(data: [u8; 35]) -> Result<Self, Self::Error> {
+    fn try_from(data: [u8; 34]) -> Result<Self, Self::Error> {
         let method =
             CloseMethod::try_from(data[0]).map_err(|e| Pay2VoutError::InvalidMethod(e.1))?;
         let address = match data[1] {
             Self::P2PKH => AddressPayload::Pkh(PubkeyHash::from_slice_unsafe(&data[2..22])),
             Self::P2SH => AddressPayload::Sh(ScriptHash::from_slice_unsafe(&data[2..22])),
             Self::P2WPKH => AddressPayload::Wpkh(WPubkeyHash::from_slice_unsafe(&data[2..22])),
-            Self::P2WSH => AddressPayload::Wsh(WScriptHash::from_slice_unsafe(&data[2..])),
+            Self::P2WSH => AddressPayload::Wsh(WScriptHash::from_slice_unsafe(&data[2..34])),
```
dr-orlovsky, line 225: The length is 34 now, so this fix is no more required

```suggestion
            Self::P2WSH => AddressPayload::Wsh(WScriptHash::from_slice_unsafe(&data[2..])),
```

dr-orlovsky (changes requested):
#210 is not wrong, it just a partial fix which makes the length of the Baid64 correct. Your PR proposes to fix the serialization of the payload which is also correct.

I have one slight comment (see below) and can merge it once rebased on master.

If you have time to add test cases like the one you put in your comment, that would be also great. Thank you

dr-orlovsky (approved):
ACK 34792e5ac2664dfaa06e3607bacbe9e28a6215d0

Thank you very much!

eauxxs:
@dr-orlovsky My pleasure. Doctor. I love rgb and hope it better 💕

Issue #213, opened 2024-06-13
RgbInvoiceBuilder with rgb20 with out date "RGB20" interface


crayon-xiao-xin:
it should be "RGB20Fixed"

dr-orlovsky:
No it should not

PR #215, opened 2024-06-13
fix: invoice builder rgb20 interface

crayon-xiao-xin:
#213


dr-orlovsky (changes requested):
No, this is wrong: invoice uses interface class name, not a specific interface

PR #219, opened 2024-06-18
add rights, attachment operations

cymqqqq:
Description: 

now, there are `fungible_ops`, and `data_ops` related to `FungibleAllocation`, and `DataAllocation`, so I think we can also have `rights_ops`, `attachment_ops` related to `RightsAllocation`, `AttachAllocation`.

1. add `rights_ops` interface
2. add `attachment_ops` interface

### DIFF
--- a/src/interface/contract.rs
+++ b/src/interface/contract.rs
@@ -443,4 +443,38 @@ impl ContractIface {
             witness_filter,
         ))
     }
+
+    pub fn rights_ops<C: StateChange<State = VoidState>>(
+        &self,
+        name: impl Into<FieldName>,
+        witness_filter: impl WitnessFilter + Copy,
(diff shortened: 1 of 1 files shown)


cymqqqq:
There is a question: I found the `info: ContractInfo` defined in the `ContractIface`, but it seems like it's not used in this struct and all the interfaces defined in the `ContractIface`.

dr-orlovsky (approved):
ACK e27ece8efffec1a064cf89d6b93f9c0d644dd2a2

dr-orlovsky:
> but it seems like it's not used in this struct and all the interfaces defined in the ContractIface.

What do you mean by "not used"? It is subject to be used by software integrating RGB: this is how it gets information about the contract like if it is on testnet, creation identity and timestamp etc.

PR #221, opened 2024-06-18
Add debug method for constructing valid consignments

### LINKED ISSUE #211: no more possible to force accepting a transfer
With beta 5 it was possible to accept a transfer with the `force` boolean, with beta 6 this possibility has been removed. @dr-orlovsky I don't recall this being communicated or discussed. This breaks our Lightning Network node so please re-introduce this ASAP (to be included in the next beta release).

dr-orlovsky:
Closes #211

### DIFF
--- a/src/containers/consignment.rs
+++ b/src/containers/consignment.rs
@@ -33,7 +33,7 @@ use amplify::{ByteArray, Bytes32};
 use armor::{ArmorHeader, AsciiArmor, StrictArmor};
 use baid64::{Baid64ParseError, DisplayBaid64, FromBaid64Str};
 use commit_verify::{CommitEncode, CommitEngine, CommitId, CommitmentId, DigestExt, Sha256};
-use rgb::validation::{ResolveWitness, Validator, Validity, Warning, CONSIGNMENT_MAX_LIBS};
+use rgb::validation::{ResolveWitness, Status, Validator, Validity, Warning, CONSIGNMENT_MAX_LIBS};
 use rgb::{
     impl_serde_baid64, validation, AttachId, BundleId, ContractHistory, ContractId, Extension,
     Genesis, GraphSeal, Operation, Schema, SchemaId, XChain,
@@ -386,6 +386,23 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
                 validation_status: status,
                 consignment: self,
             })
+        }
+    }
+
+    /// Method to construct valid consignment which can be consumed by a stash
+    /// for debug purposes.
+    ///
+    /// # Safety
+    ///
+    /// The function runs no actual validation, but still constructs
+    /// [`ValidConsignment`] type, which is a violation of type safety.
+    pub unsafe fn assume_valid(self) -> ValidConsignment<TRANSFER> {
+        let validation_status = Status::with_failure(validation::Failure::Custom(s!(
+            "consignment was assumed valid with no actual validation"
+        )));
+        ValidConsignment {
+            validation_status,
+            consignment: self,
         }
     }
 }
--- a/src/containers/consignment.rs
+++ b/src/containers/consignment.rs
@@ -396,7 +396,8 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
     ///
     /// The function runs no actual validation, but still constructs
     /// [`ValidConsignment`] type, which is a violation of type safety.
-    pub unsafe fn assume_valid(self) -> ValidConsignment<TRANSFER> {
+    /// The method must be used only for tests.
+    pub fn assume_valid_unsafe(self) -> ValidConsignment<TRANSFER> {
         let validation_status = Status::with_failure(validation::Failure::Custom(s!(
             "consignment was assumed valid with no actual validation"
         )));
--- a/src/containers/consignment.rs
+++ b/src/containers/consignment.rs
@@ -389,14 +389,14 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
         }
     }
 
-    /// Method to construct valid consignment which can be consumed by a stash
-    /// for debug purposes.
+    /// Method to forcefully construct a valid consignment, which can be consumed by a stash.
     ///
     /// # Safety
     ///
-    /// The function runs no actual validation, but still constructs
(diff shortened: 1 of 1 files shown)


nicbus (changes requested):
I see the method is documented as "for debug purposes". I don't think this should be a "debug" method, as it's going to be used in "production" code, like accepting the consignment for a funding transaction when opening an RGB LN channel. Having it documented as "debug" makes me think it can be easily dropped in the future without worrying about code that might use it, similar to what happened to the `force` parameter in `accept_transfer`. I think this needs to be documented as a method that's intended for special cases and should not be used if you don't know what you're doing, not a debug one.

The method is defined as `unsafe`, but it doesn't seem unsafe in the rust sense, so I think it should be named unsafe (e.g. assume_valid_unsafe) but not defined as rust unsafe, as that's not the case and it forces users to wrap calls into an unneeded unsafe block.

The comment says this violates type safety. I get the general idea, but what do you mean exactly?

I'm not sure about the custom validation failure. Having a custom warning or info makes sense to me, but a failure feels wrong, as the whole purpose of the method is to render the consignment valid, while a failure indicates validation failed.

dr-orlovsky:
What you say makes perfect sense if this method is required for normal lightning functioning. But I do not see the scenario whether it is required. Can you please give me more information why a node must accept consignment for lightning channel funding without a validation?

The type safety is the fact that given type provides certain guarantees on its data at runtime which are ensured by the compiler at compile-time (here it is the concept of the validation) - meaning that the method is `unsafe` in pure rust sense, since it allows to bypass such validation.

nicbus:
> What you say makes perfect sense if this method is required for normal lightning functioning. But I do not see the scenario whether it is required. Can you please give me more information why a node must accept consignment for lightning channel funding without a validation?

A node needs to accept the funding transaction consignment (which allocates RGB assets to the multisig output) before it's broadcast, in order to generate the first pair of commitment transactions (spending the funding and allocating the RGB assets to the local and remote outputs).

To be precise, the requirement is not that the consignment is not validated. It just needs to be accepted even if the witness transaction is still missing from the blockchain, as that's expected.

> The type safety is the fact that given type provides certain guarantees on its data at runtime which are ensured by the compiler at compile-time (here it is the concept of the validation) - meaning that the method is unsafe in pure rust sense, since it allows to bypass such validation and breaks compile-time type guarantees. Thus, unsafe is required to notify a user about that happening at compile-time.

I'm not sure I follow you here. There doesn't seem to be anything that the compiler is checking that would fail if validation is skipped. Instead, it's business logic that's not run when validation is not executed, but this is not at the rust compiler level. To prove that to myself, I dropped the `unsafe` keyword from the `assume_valid` method definition and it still compiles fine.

I get your point, you're trying to protect users from unwanted behavior and risks, and I support that, but I don't see the need to use a rust feature in an improper way to drive the point home. Naming and documentation are in my opinion the correct choice here.

Moreover, as said above, it's not even the case that this method should do anything "wrong". The need here is to be able to accept a (valid) consignment before the witness transaction has been broadcast. In the context of a funding transaction and a pair of commitment transactions there is no issue, as there's no point in broadcasting a commitment if the funding has never been broadcast itself. On the other hand, broadcasting the funding before constructing the first pair of commitments would be a serious issue.

dr-orlovsky:
> To be precise, the requirement is not that the consignment is not validated. It just needs to be accepted even if the witness transaction is still missing from the blockchain, as that's expected.

Well, we have discussed that: you just need to report the witness transaction as valid: this is the task which must be handled by the lightning node resolver. No need in force-converting anything.

> There doesn't seem to be anything that the compiler is checking that would fail if validation is skipped. <...> To prove that to myself, I dropped the unsafe keyword from the assume_valid method definition and it still compiles fine.

Sorry, now I can't understand what you are saying. Of course if you remove `unsafe` it will compile fine, that's why `unsafe` keyword is needed to ensure that compiler notifies about type safety violation!

dr-orlovsky:
For the history record, I am strongly against having and using this method, since it violates the security model.

Details: https://github.com/RGB-WG/rgb-std/issues/211#issuecomment-2194581941

Strong conceptual NACK on merging this PR

nicbus:
@dr-orlovsky I have pushed a commit on top of yours, to rephrase the docstring and add stronger warnings against the usage of this method. Please let me know if you think this is enough or, if you think even stronger language should be used, what you would add, so I can further improve it.

src/containers/consignment.rs
```diff
@@ -389,14 +389,14 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
         }
     }
 
-    /// Method to construct valid consignment which can be consumed by a stash
-    /// for debug purposes.
+    /// Method to forcefully construct a valid consignment, which can be consumed by a stash.
     ///
```
dr-orlovsky, line 393: Sorry, I strictly disagree with this. It has - and must not have (!) no other use than debug purposes, since any other use - is security leak. Literally, my opinion is so strong, that until the public audit conclusion on this issue I will just fork this repository and will contribute to the fork not having this security issue.

src/containers/consignment.rs
```diff
@@ -389,14 +389,14 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
...
     /// # Safety
     ///
-    /// The function runs no actual validation, but still constructs
-    /// [`ValidConsignment`] type, which is a violation of type safety.
-    /// The method must be used only for tests.
+    /// Warning: this runs no actual validation, but still constructs a [`ValidConsignment`]. If
+    /// used improperly this could lead to loss of funds. The method must be used only for special
+    /// purposes where it's necessary to consume a consignment even though standard rules prevent
```
dr-orlovsky, line 398: There are no other "special purposes" than debugging!

dr-orlovsky (commented):
Strong NACK

nicbus:
We had a misunderstanding. I took you giving me permission to merge as an ACK but it's then become clear that that was not the case. It was never my intention to merge this against your will and reasoning. Quite the opposite, the idea was to cooperate to find the best solution.

Since then, we had discussions elsewhere and it has all been clarified. This solution is not good for production code, so it's not what we desire either.

A better alternative has been found in https://github.com/RGB-WG/rgb-std/pull/233, which makes this PR no more needed, so I'm closing it.

PR #222, opened 2024-06-19
add interface for contract info

cymqqqq:
Description: there is an `info: ContractInfo` in the ContractIface, in this pull request, the following contract info interfaces are added to get related contract information.

1. Schema ID in string format
2. issue identity in string format
3. issue date time 
4. is on the testnet or not
5. alt layer set

So the developers can directly call the related contract interface to get contract info(such as `self.info.issuer_identity()` to get the contract issuer identity in string format).

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -19,13 +19,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap};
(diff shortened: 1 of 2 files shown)


dr-orlovsky (commented):
Sorry, I do not understand why this is needed. The fields are `pub` anyway, you do not need methods to access them.

Concept NACK

cymqqqq:
Ok, I will close this PR.

PR #223, opened 2024-06-19
warn about repeated field name in iimpl

cymqqqq:
Description: warn about repeated field names in the iface impl.

1. new `RepeatWarning` enum type for warning about the repeated field name and counts.
2. new `RepeatedStatus` for adding the repeated warnings into a vector.
4. use hashmap to check the repeated times in the metadata, global state, valencies, transitions, and extensions.
5. modified error, now is `(RepeatedStatus, Vec<ImplInconsistency>)`

### DIFF
--- a/src/interface/iimpl.rs
+++ b/src/interface/iimpl.rs
@@ -567,7 +567,7 @@ impl IfaceImpl {
                 errors.push(ImplInconsistency::SchemaExtensionAbsent(field.name.clone(), field.id));
             }
         }
-        
+
         dup_extensions.iter().for_each(|(field_name, &count)| {
             if count > 1 {
                 status.add_warning(RepeatWarning::RepeatedExtensions(field_name.clone(), count));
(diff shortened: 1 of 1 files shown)


dr-orlovsky (changes requested):
It is not clear to me why we need to introduce new types instead of adding messages to existing `Vec<ImplInconsistency>` type

src/interface/iimpl.rs
```diff
@@ -393,76 +418,134 @@ impl IfaceImpl {
...
             if !schema.meta_types.contains_key(&field.id) {
                 errors.push(ImplInconsistency::SchemaMetaAbsent(field.name.clone(), field.id));
             }
         }
 
+        dup_metadata.iter().for_each(|(field_name, &count)| {
+            if count > 1 {
+                errors.push(ImplInconsistency::RepeatedMetaData(field_name.clone(), count));
+            }
+        });
```
dr-orlovsky, line 434: This (and below) can be expressed more "rusty":
```suggestion
        dup_metadata.iter().filter(|(_, &count)| count > 1).for_each(|(field_name, &count)| {
            errors.push(ImplInconsistency::RepeatedMetaData(field_name.clone(), count));
        });
```

dr-orlovsky (changes requested):
One small improvement, the rest looks fine for me

dr-orlovsky (approved):
ACK d7eac9ae1d98a71414649985309178959c2dc759

PR #225, opened 2024-06-22
Check attach

cymqqqq:
Description: check attached ids from data containers are present in the consignment.

1. iter bundle ids from index consignment
2. index transition bundle by bundle id
3. for each transition bundle, get all transitions
4. for each transition, get all assignments, and then get the revealed attach state by `as_attachment()`
5. finally, check attached id from the consignment is equal to attached id from the reveal attach state(also, from the operation transition attach state).

I'm not sure if the logic I implemented is correct.

### DIFF
--- a/src/containers/consignment.rs
+++ b/src/containers/consignment.rs
@@ -384,11 +384,7 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
                 .known_transitions
                 .values()
                 .for_each(|transition| {
-                    transition.assignments.values().for_each(|assign| {
-                        assign.as_attachment().iter().for_each(|item| {
-                            let state = item
-                                .as_revealed_state()
-                                .expect("get revealed attached state failed");
+                    transition.assignments.values().filter_map(|assign| assign.as_attachment().and_then(Attach::as_revealed_state)).for_each(|assign| {
                             if !self.attachments.keys().any(|&id| id == state.id) {
                                 status.add_warning(Warning::Custom(format!(
                                     "attach id from data containers {:?} is not present in the \
--- a/src/containers/consignment.rs
+++ b/src/containers/consignment.rs
@@ -376,25 +376,31 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
                 )));
             }
         }
+        index.bundle_ids().for_each(|bundle_id| {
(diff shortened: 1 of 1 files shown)


src/containers/consignment.rs
```diff
@@ -376,6 +376,31 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
...
+                                     consignment",
+                                    state.id
+                                )));
+                            }
+                        })
+                    })
+                })
+        });
+
         // TODO: check attach ids from data containers are present in operations
```
dr-orlovsky, line 404: We can remove this line now

src/containers/consignment.rs
```diff
@@ -376,6 +376,31 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
...
+                .expect("index transition bundle failed");
+            transition_bundle
+                .known_transitions
+                .values()
+                .for_each(|transition| {
+                    transition.assignments.values().for_each(|assign| {
+                        assign.as_attachment().iter().for_each(|item| {
+                            let state = item
+                                .as_revealed_state()
+                                .expect("get revealed attached state failed");
```
dr-orlovsky, line 391:
```suggestion
                    transition.assignments.values().filter_map(|assign| assign.as_attachment().and_then(Attach::as_revealed_state)).for_each(|assign| {
```

src/containers/consignment.rs
```diff
@@ -376,6 +376,31 @@ impl<const TRANSFER: bool> Consignment<TRANSFER> {
...
+                            if !self.attachments.keys().any(|&id| id == state.id) {
```
dr-orlovsky, line 392:
```suggestion
                            if !self.contains(state.id) {
```

dr-orlovsky (changes requested):
Few improvements

src/containers/consignment.rs
cymqqqq: There is a question: it's not clear to me why `self.contains`, because `self` is for the consignment, and no `contains` method here.

dr-orlovsky: Sorry it should be `self.attachments.contains_key`

cymqqqq:
I will close this pr, because I have no good idea for `check attach` now.

PR #227, opened 2024-06-25
add statechange for owned-fraction in rgb21 invoice data

cymqqqq:
Background: 

Now there are `Amount`, and `Ownedfraction` in the RGB invoice module, that is 
`pub struct Amount( u64);` and `pub struct OwnedFraction(u64);`.  We can perform a state change operation for the amount type when we want to build a `transfer` transition, and there is also the `check_add` method for the amount type.
Both the `StateChange` and `check_add` for the amount type can perform algorithm operation(+, -). 

Then for the rgb21 invoice, there is Ownedfraction which is `u64` similar to `Amount`, and now if we want to build a `transfer` operation for the rgb21 contract, we just push all of them(owned fraction, token index) into the data state. So I think maybe there is also a state change and a `check_add` method requirement for the `Ownedfraction` type, and then the user can add an owned fraction when they want to transfer. Finally, they can push (changed owned fraction, token index) into a data state.

Description:

impl StateChange for Ownedfraction


src/interface/contract.rs
```diff
@@ -121,6 +122,58 @@ pub trait StateChange: Clone + Eq + StrictDumb + StrictEncode + StrictDecode {
...
+
+    fn merge_spent(&mut self, sub: Self::State) {
+        *self = match self {
+            OwnedFractionChange::Dec(neg) => OwnedFractionChange::Dec(*neg + sub),
+            OwnedFractionChange::Zero => OwnedFractionChange::Dec(sub),
+            OwnedFractionChange::Inc(pos) if *pos > sub => OwnedFractionChange::Inc(*pos - sub),
+            OwnedFractionChange::Inc(pos) if *pos == sub => OwnedFractionChange::Zero,
+            OwnedFractionChange::Inc(pos) if *pos < sub => OwnedFractionChange::Dec(sub - *pos),
+            OwnedFractionChange::Inc(_) => unreachable!(),
+        };
```
dr-orlovsky, line 162: with `Ord` you can get rid of `unreachable` and instead enjoy compiler-provided exhaustive security for matches:

```suggestion
        *self = match self {
            OwnedFractionChange::Dec(neg) => OwnedFractionChange::Dec(*neg + sub),
            OwnedFractionChange::Zero => OwnedFractionChange::Dec(sub),
            OwnedFractionChange::Inc(pos) => match sub.cmp(pos) {
              Ordering::Less => OwnedFractionChange::Inc(*pos - sub),
              Ordering::Equal => OwnedFractionChange::Zero,
              Ordering::Greater => OwnedFractionChange::Dec(sub - *pos),
            },
        };
```

src/interface/contract.rs
```diff
@@ -121,6 +122,58 @@ pub trait StateChange: Clone + Eq + StrictDumb + StrictEncode + StrictDecode {
...
+    }
+
+    fn merge_received(&mut self, add: Self::State) {
+        *self = match self {
```
dr-orlovsky, line 166: Ibid

dr-orlovsky (changes requested):
I think it is always nice to have compiler guarantees instead of `unreachable` - pls see my comments

src/interface/contract.rs
```diff
@@ -155,21 +156,23 @@ impl StateChange for OwnedFractionChange {
...
     fn merge_received(&mut self, add: Self::State) {
         *self = match self {
             OwnedFractionChange::Inc(pos) => OwnedFractionChange::Inc(*pos + add),
             OwnedFractionChange::Zero => OwnedFractionChange::Inc(add),
-            OwnedFractionChange::Dec(neg) if *neg > add => OwnedFractionChange::Dec(*neg - add),
-            OwnedFractionChange::Dec(neg) if *neg == add => OwnedFractionChange::Zero,
-            OwnedFractionChange::Dec(neg) if *neg < add => OwnedFractionChange::Inc(add - *neg),
-            OwnedFractionChange::Dec(_) => unreachable!(),
+            OwnedFractionChange::Dec(neg) => match add.cmp(neg) {
+                Ordering::Greater => OwnedFractionChange::Dec(*neg - add),
```
dr-orlovsky, line 172: This must be less, not greater - the order of comparison is reversed in the match

dr-orlovsky (changes requested):

dr-orlovsky (approved):
ACK. Will wait for review of those actually working with NFTs - @zoedberg and @crisdut

dr-orlovsky:
The PR makes sense for me. @zoedberg @crisdut can you confirm the same so we can merge it before the next release?

dr-orlovsky:
Not anymore applicable in the context of the new architecture

PR #228, opened 2024-06-26
unreachable to ord for amount

cymqqqq:
Background:

Based on https://github.com/RGB-WG/rgb-std/pull/227, there is also an `unreachable` in the rgb20 amount, so I think maybe we can change to the `Ord` in order to get rid of unreachable.

Description:

remove the `unreachable` in the rgb20 amount state change, then use the `Ord`. Similar to the rgb21 owed fraction state change.

### DIFF
--- a/src/interface/contract.rs
+++ b/src/interface/contract.rs
@@ -168,9 +168,9 @@ impl StateChange for AmountChange {
             AmountChange::Inc(pos) => AmountChange::Inc(*pos + add),
             AmountChange::Zero => AmountChange::Inc(add),
             AmountChange::Dec(neg) => match add.cmp(neg) {
-                Ordering::Greater => AmountChange::Dec(*neg - add),
+                Ordering::Less => AmountChange::Dec(*neg - add),
                 Ordering::Equal => AmountChange::Zero,
-                Ordering::Less => AmountChange::Inc(add - *neg),
+                Ordering::Greater => AmountChange::Inc(add - *neg),
             },
         };
     }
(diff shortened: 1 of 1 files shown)


cryptoquick (approved):
Very nice

src/interface/contract.rs
```diff
@@ -154,21 +155,23 @@ impl StateChange for AmountChange {
...
     fn merge_received(&mut self, add: Self::State) {
         *self = match self {
             AmountChange::Inc(pos) => AmountChange::Inc(*pos + add),
             AmountChange::Zero => AmountChange::Inc(add),
-            AmountChange::Dec(neg) if *neg > add => AmountChange::Dec(*neg - add),
-            AmountChange::Dec(neg) if *neg == add => AmountChange::Zero,
-            AmountChange::Dec(neg) if *neg < add => AmountChange::Inc(add - *neg),
-            AmountChange::Dec(_) => unreachable!(),
+            AmountChange::Dec(neg) => match add.cmp(neg) {
+                Ordering::Greater => AmountChange::Dec(*neg - add),
```
dr-orlovsky, line 171: This must be less, not greater, since we reverse the order `neg` and `add` are compared

```suggestion
                Ordering::Less => AmountChange::Dec(*neg - add),
```

dr-orlovsky (changes requested):

dr-orlovsky (approved):

dr-orlovsky:
I already asked few times to use rebases instead of useless multiple merges of `master` branch all over the place. Unless there will be a normal commitment history - no `master` merges, one commit per logical change - I will be just sqash-merging your commits.

PR #229, opened 2024-06-26
check add and sub operation for owned fraction

cymqqqq:
Background:
In the rgb21 contract, we have `error fractionOverflow
		"The amount of token fractions in outputs exceeds 1"`, so I think maybe we should have a check add/sub-operation to check if or not the owned fraction is overflow(or the other fraction checks for add/sub-operation), also we can make a change for the allocation. It's similar to the rgb20 amount.

Description:

add `check_add`, `check_sub`, `check_addassign`, `check_subassign`, and other methods for the `Ownedfraction`.


cryptoquick (approved):
Love the tests... It's also funny to see codecov flag tests as untested

dr-orlovsky (approved):
ACK 7f637c6b5735bee4c8eff98fa3586f025641b7b5

PR #230, opened 2024-06-26
Collect supplyment features in consign

cymqqqq:
Description:

1. add a `get_sigs` interface for obtaining the content_sig in order to collect signatures. (the reason for adding this interface is that there are `add_supplyment`, and `supplement` interfaces for adding and obtaining the supplement, and then I noticed that there is an `import_sigs` interface, but no method to get content signature by content ID, so I think maybe we need an interface to obtain the content signature)

2. collect supplements.
3. test cases.

### DIFF
--- a/src/persistence/memory.rs
+++ b/src/persistence/memory.rs
@@ -173,6 +173,10 @@ impl StashReadProvider for MemStash {
             .into_iter())
     }
 
+    fn get_sigs(&self, content_id: &ContentId) -> Result<Option<&ContentSigs>, Self::Error> {
+        Ok(self.sigs.get(&content_id))
+    }
+
     fn geneses(&self) -> Result<impl Iterator<Item = &Genesis>, Self::Error> {
         Ok(self.geneses.values())
     }
--- a/src/persistence/memory.rs
+++ b/src/persistence/memory.rs
@@ -174,7 +174,7 @@ impl StashReadProvider for MemStash {
     }
 
     fn get_sigs(&self, content_id: &ContentId) -> Result<Option<&ContentSigs>, Self::Error> {
-        Ok(self.sigs.get(&content_id))
+        Ok(self.sigs.get(content_id))
     }
 
     fn geneses(&self) -> Result<impl Iterator<Item = &Genesis>, Self::Error> {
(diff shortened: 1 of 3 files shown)


src/persistence/memory.rs
```diff
@@ -173,6 +173,10 @@ impl StashReadProvider for MemStash {
             .into_iter())
     }
 
+    fn get_sigs(&self, content_id: &ContentId) -> Result<Option<&ContentSigs>, Self::Error> {
```
dr-orlovsky, line 176:
```suggestion
    fn sigs_for(&self, content_id: &ContentId) -> Result<Option<&ContentSigs>, Self::Error> {
```

src/persistence/stock.rs
```diff
@@ -741,32 +740,21 @@ impl<S: StashProvider, H: StateProvider, P: IndexProvider> Stock<S, H, P> {
...
-                    supplyments.insert(iface_suppl.clone());
-                    supplyments.insert(iimpl_suppl.clone())
-                }
-                (Some(iface_suppl), None) => supplyments.insert(iface_suppl.clone()),
-                (None, Some(iimpl_suppl)) => supplyments.insert(iimpl_suppl.clone()),
-                (None, None) => false,
-            };
+            self.stash
+                .supplement(ContentRef::Iface(iface.iface_id()))?
+                .and_then(|iface_suppl| Some(supplements.insert(iface_suppl.clone())));
```
dr-orlovsky, line 754: Here and avove, actually, this should be `map`:

```suggestion
                .map(|iface_suppl| supplements.insert(iface_suppl.clone()));
```

src/persistence/stock.rs
```diff
@@ -791,8 +779,8 @@ impl<S: StashProvider, H: StateProvider, P: IndexProvider> Stock<S, H, P> {
 
         let (types, scripts) = self.stash.extract(&schema_ifaces.schema, ifaces.keys())?;
         let scripts = Confined::from_iter_unsafe(scripts.into_values());
-        let supplyments: Confined<BTreeSet<Supplement>, ZERO, U8> =
-            Confined::try_from(supplyments).map_err(|_| ConsignError::TooManySupplies)?;
+        let supplements: Confined<BTreeSet<Supplement>, ZERO, U8> =
```
dr-orlovsky, line 782:
```suggestion
        let supplements =
```

dr-orlovsky (changes requested):

dr-orlovsky (approved):

PR #231, opened 2024-06-27
collect supplements in issue contract

cymqqqq:
Description:

add collect supplements feature for the `issue_contract`, includes the following Supplement instance:

genesis supplement
iface supplement
iimpl supplement
schema supplement

### DIFF
--- a/src/interface/builder.rs
+++ b/src/interface/builder.rs
@@ -21,9 +21,9 @@
 
 #![allow(clippy::result_large_err)]
 
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 
-use amplify::confinement::{Confined, SmallOrdSet, TinyOrdMap, U16};
+use amplify::confinement::{Confined, SmallOrdSet, TinyOrdMap, U16, U8, ZERO};
 use amplify::{confinement, Wrapper};
 use chrono::Utc;
 use invoice::{Allocation, Amount};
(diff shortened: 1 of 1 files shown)


dr-orlovsky (changes requested):
You can just do
```
supplements: confined_bset![
   Supplement::new(ContentRef::Schema(schema.schema_id()), schema.developer.clone()),
    ...
]
```

cymqqqq:
Sorry....I always use the `update branch` button, next time I'll try to `git rebase main`.

dr-orlovsky (changes requested):
Sorry, just understood that in this PR instead of gathering supplements from the stash you create some pseudo-supplements out of arbitrary cosnensus data.

Approach NACK. Please either collect supplements from the stash - or close the PR

cymqqqq:
I agree with you, the first idea is to collect supplements from the stash, but the question is: do we have a stash when we issue a contract? (In the `consign` method, yes, there is already a stash so we can collect supplements from the stash.) I just found `ContractBuilder`, and after we finish the contract issue, then initialize a stock(`MemStash`, `MemState`, and `MemIndex`), and then we can import the kit and contract to the stash.

dr-orlovsky:
When you issue contract you need to collect just supplements for interfaces, schema and implementation. There is no "genesis supplement" which can be found in the stash.

You also need to keep in mind that right now there is no supplements which exist - this functionality is there only as a way to add new things in the future.

PR #234, opened 2024-07-06
Refactor FS store and load workflows

dr-orlovsky:
Counterpart to https://github.com/BP-WG/bp-wallet/pull/44

Provides more efficient and clearer way to ensure that the data are always persisted to disk when needed.


zoedberg (approved):
tACK c2f83473eb612638b9b67c0fb5150c574c35171e

PR #237, opened 2024-07-12
Support Testnet4 Network

will-bitlight:
This PR updates rgb-std to be compatible with the [latest Network changes](https://github.com/BP-WG/bp-std/pull/32), including support for Testnet4. 

It also fixes a compilation error encountered during the development process.

If you have any suggestions, please leave your valuable comments, I will actively cooperate with the revision, thank you very much

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,8 +280,7 @@ dependencies = [
 [[package]]
 name = "bp-invoice"
 version = "0.11.0-beta.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


dr-orlovsky (changes requested):
We need to remove `mut`, not the reference. Also, the patch should go into a dedicated section of Cargo.toml as I pointed out elsewhere

will-bitlight:
> We need to remove `mut`, not the reference. Also, the patch should go into a dedicated section of Cargo.toml as I pointed out elsewhere

I agree with you. We're done :)

dr-orlovsky (approved):

PR #251, opened 2024-08-09
Add store providers to persistence module

dr-orlovsky:
This solves the problem with saving and restoring Stock and related persistence objects in a custom way.

If a client of the library want to use custom store and load procedures, it has to do the following:
- create a custom type with all necessary information;
- implement `StoreProvider` for with `MemStash`, `MemState` and `MemIndex` generics
- implement `StockStoreProvider` for it

You can check how it is done in https://github.com/RGB-WG/rgb-std/compare/store?expand=1#diff-ad8cec98dd4a1dac3d2a20e6c7a9c3e64129a3182285b50e1e3df20da2c167c7 where `PathBuf` becomes such store provider for a local file. If somebody needs to use Amazon S3 for instance, they can wrap this `PathBuf` into a custom structure and use its methods when needed.

Two other related PRs:
- https://github.com/BP-WG/bp-wallet/pull/66
- https://github.com/RGB-WG/rgb/pull/238


will-bitlight:
Dear Maxim, I have a colleague who is paying attention to this PR at any time and cooperating with the docking work of the client. You can replace him with @Crayon-Shin-chan-1024  reviewer, so that your design can have a better understanding with the downstream projects :)

dr-orlovsky:
Thank you for the suggestion! First, he needs to accept invitation to the org

PR #255, opened 2024-08-12
Iignore the difference in state and seals of output assignments

### LINKED ISSUE #238: How to properly handle ephemeral state in state channels
I've added a test here https://github.com/zoedberg/rgb-integration-tests/tree/ln_transfers that reproduces an issue we encountered on our LN node (run `cargo test --test transfers ln_htlc_transfer`).

In the test we try to simulate the creation of a commitment TX and of an HTLC TX spending a vout of the commitment. In this case RGB fails to find inputs when calling the `contract_assignments_for` method. I assume this is because we haven't consumed the fascia. But AFAIR it would be incorrect to consume the fascia of a commitment TX that hasn't been broadcast and might never be. Therefore I'm not sure how we should address this, I normally call `contract_assignments_for` to obtain the `Opout` and `PersistedState` needed to call the `add_input` method of the `TransitionBuilder`. 

Just a note: to mimic the behavior of LN we had to add some custom methods to create and color a PSBT, this code is mostly copied from RGB repos so we expect it to be correct. We will extend these methods to make also some complex validation tests but for now it should be sufficient to demonstrate the problem.

dr-orlovsky:
Required for supporting LN channels. Closes #238

### DIFF
--- a/src/contract/assignments.rs
+++ b/src/contract/assignments.rs
@@ -67,20 +67,10 @@ pub struct OutputAssignment<State: KnownState> {
 
 impl<State: KnownState> PartialEq for OutputAssignment<State> {
     fn eq(&self, other: &Self) -> bool {
-        if self.opout == other.opout &&
-            (self.seal != other.seal ||
-                self.witness != other.witness ||
-                self.state != other.state)
-        {
-            panic!(
-                "RGB was provided with an updated operation using different witness transaction. \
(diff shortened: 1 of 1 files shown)


zoedberg (approved):
tACK c6358c24d5522a206b8d783c94d2d8a1ccb1116c

PR #264, opened 2024-08-30
Fix invalid interface abstraction in MemStash::impl_for

### LINKED ISSUE #263: Interface resolves to an "abstract" interface out of inheritance hierarchy
In the interface wrappers there are panics raised when trying to read a contract with the incorrect interface (e.g. https://github.com/RGB-WG/rgb-interfaces/blob/master/src/rgb20/wrapper.rs#L86). These panics have broken a method in rgb-lib, where we get a contract ID and a contract interface and we verify that the user has provided the correct iface (e.g. doing `let contract = runtime.contract_iface_class::<Rgb20>(contract_id);`). @dr-orlovsky please suggest a workaround for this or remove the panic

dr-orlovsky:
Closes #263

### DIFF
--- a/src/persistence/memory.rs
+++ b/src/persistence/memory.rs
@@ -192,8 +192,8 @@ impl StashReadProvider for MemStash {
             .values()
             .find(|iimpl| C::IFACE_IDS.contains(&iimpl.iface_id))
             .or_else(|| {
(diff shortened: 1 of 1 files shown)


zoedberg (approved):
tACK bd38d2258b466b69fc20513a9d5b25c71724f4c4

PR #265, opened 2024-09-02
Require just one secret seal for the transfer method

### LINKED ISSUE #262: Sending same asset to multiple recipients with a single TX
This commit https://github.com/RGB-WG/rgb-std/commit/bc2ec1c22a4bfe3e4b394b9f133a6ef9cb6082a3 has broken the possibility of sending the same asset to multiple recipients with a single TX.

In specific here:
```diff
-    /// Set of seals which are history terminals.
-    pub terminals: SmallOrdMap<BundleId, Terminal>,
+    /// Set of secret seals which are history terminals.
+    pub terminals: SmallOrdMap<BundleId, XChain<SecretSeal>>,
```
`terminals` lost support for multiple `XChain<SecretSeal>`. In the parent commit the [`Terminal` struct](https://github.com/RGB-WG/rgb-std/blob/5b116661704b1cfcb8a3d2042d89b50ff264615a/src/containers/util.rs#L54) was:
```rust
pub struct Terminal {
    pub seals: SmallOrdSet<XChain<TerminalSeal>>,
}
```
which means that multiple seals under the same bundle where supported.

This was working in `0.10`, then it was not working in `0.11.0-beta.5` (not sure from which commit it stopped working), it has been fixed by @eauxxs in https://github.com/RGB-WG/rgb-std/pull/200 (and therefore it was working in `0.11.0-beta.6`) and then it has been broken again with https://github.com/RGB-WG/rgb-std/commit/bc2ec1c22a4bfe3e4b394b9f133a6ef9cb6082a3 (and therefore is broken in `0.11.0-beta.7`).

dr-orlovsky:
Closes #262

Requires https://github.com/RGB-WG/rgb/pull/240

### DIFF
--- a/src/persistence/stock.rs
+++ b/src/persistence/stock.rs
@@ -708,7 +708,7 @@ impl<S: StashProvider, H: StateProvider, P: IndexProvider> Stock<S, H, P> {
         &self,
         contract_id: ContractId,
     ) -> Result<Contract, StockError<S, H, P, ConsignError>> {
-        let consignment = self.consign::<false>(contract_id, [], [])?;
+        let consignment = self.consign::<false>(contract_id, [], None)?;
         Ok(consignment)
     }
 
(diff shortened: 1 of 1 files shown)


zoedberg (approved):
ACK 526c2668c41dd0386b0df2be5715cc6d70a9fa9a

PR #272, opened 2024-09-17
Always add opret anchors when preset to a consignment

### LINKED ISSUE #271: receiver fails to accept a transfer which spends opret and tapret allocations together
As shown by the `tapret_wlt_receiving_opret` test in https://github.com/RGB-WG/rgb-integration-tests/pull/6 there's an issue regarding the spend of opret and tapret allocations together. The sender is able to create a consignment but then the receiver fails to accept it.

dr-orlovsky:
Closes #271

I believe it should fix https://github.com/RGB-WG/rgb-tests/pull/6


zoedberg (approved):
I've run the test in https://github.com/RGB-WG/rgb-tests/pull/6 100 times and can confirm the test passed every time, so with this PR there is no more a non-deterministic behavior and transfers spending both tapret and opret allocations work again.

PR #283, opened 2024-11-01
Remove consignment-based resolver

### LINKED ISSUE #280: cannot distinguish allocations coming from unsafe history
As shown in RGB-WG/rgb-tests#22 a wallet is unable to distinguish an allocation coming from an "unsafe" history (i.e. a history containing unbroadcasted/unmined TXs) from allocations coming from a "safe" history. I think either these kind of allocations should be distinguished or the wallet should not be able to accept a transfer with unsafe history when using a "normal" resolver (i.e. one using an indexer).

dr-orlovsky:
Closes #280


zoedberg (approved):
ACK

PR #285, opened 2024-11-25
fix nightly build and lint

zoedberg:
This PR fixes the build on nightly. @dr-orlovsky I have used 0.11 beta 9 as base for the commit and would be glad if you could release a beta 9.1 for this (it would fix the new CI job added in https://github.com/RGB-WG/rgb-tests/pull/25)


dr-orlovsky (approved):
ACK 62b48da94eb139cf0793596becac7d208a38e2db

PR #286, opened 2024-12-10
RGB v0.12: zk-STARKy, fully refactored and simplified

dr-orlovsky:


dr-orlovsky:
Released as v0.12.0-beta.4

src/popls/bp.rs
```diff
@@ -278,31 +278,37 @@ impl OpRequest<Option<WoutAssignment>> {
...
+                .entry(&out.name)
+                .or_insert_with(|| api.calculate(out.name.clone()));
             calc.lessen(&out.state.data)?;
         }
-        if !calc.diff()?.is_empty() {
-            Err(UnmatchedState::NotEnoughChange)
-        } else {
-            Ok(calc)
+        for (state_name, calc) in calcs {
+            if !calc.diff()?.is_empty() {
```
crisdut, line 308: I believe that verification fails if the contract is fungible.

According to the current [implementation](https://github.com/AluVM/sonic/blob/v0.12.0-beta.4/api/src/adaptors/embedded.rs#L280-L284), even if the sum of the outputs minus the inputs was zero, the diff method will return a vector that contains one element.

PR #291, opened 2025-01-06
Use XOupoint instead of XOutputSeal in the APIs

### LINKED ISSUE #290: Diverging opret/tapret assignments are skipped for an outpoint
The existing API which allows to list contracts and specific assignments to a given bitcoin tx outpoint works only withing a scope of given single-use seal method, since it uses `XOutputSeal` (which includes closing method information) instead of `XOutpoint` (which doesn't). Thus, when you have diverging tapret/opret assignments to the same outpoint, one type of them is always ignored, which leads to a loss of assets.

Test demonstrating the issue: https://github.com/RGB-WG/rgb-tests/pull/24 (`NoInput` variant of the failure is caused by this problem)

dr-orlovsky:
Includes fix from #289 
Closes #290

### DIFF
--- a/src/containers/partials.rs
+++ b/src/containers/partials.rs
@@ -156,12 +156,16 @@ impl Hash for TransitionInfo {
 }
 
 impl TransitionInfo {
+    /// # Panics
+    ///
+    /// If the number of provided seals is zero.
     pub fn new(
         transition: Transition,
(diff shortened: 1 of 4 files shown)


zoedberg (approved):
tACK af5a31bf8ee8ffc879ec358f31e976338739e02c

PR #293, opened 2025-01-14
Unify single-use seals

dr-orlovsky:


dr-orlovsky:
It was actually merge, but github is just being stupid

PR #299, opened 2025-03-02
feat(mound): add contract articles import functionality

astral-bitlight:
This PR introduces a new `import_articles` function to the `DirMound` implementation, allowing for the import of contract articles from external files.

### DIFF
--- a/src/mound.rs
+++ b/src/mound.rs
@@ -470,21 +470,22 @@ pub mod file {
                 let codex_id = schema.codex.codex_id();
                 let name = articles.contract.meta.name.to_string();
                 let mut old_stock = Stock::load(src_path);
-                let stock_export_tmpfile = self.path().join(format!("{}.tmp", name));
-                old_stock.backup_to_file(&stock_export_tmpfile).map_err(|err| {
+                let writer = StrictWriter::in_memory();
+                old_stock.export_all(writer).map_err(|err| {
                     io::Error::new(
                         io::ErrorKind::InvalidData,
                         format!("Failed to backup contract: {}", err),
                     )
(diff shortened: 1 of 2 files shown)


astral-bitlight:
We need sonic to expose path of FileSupply  to prevent manually concat `NAME.contract` dir:

```
    impl FileSupply {
        pub fn path(&self) -> &Path {
            &self.path
        }
```

-->

```
            // Persist articles to target directory
            let new_supply = FileSupply::new(&name, &target_dir);
            let supply_dir = new_supply.path().to_owned(); // here
            let _new_pile = FilePile::<SealDef>::new(&name, &target_dir);
            let _new_stock = Stock::create(articles, new_supply);

            let stockpile = Stockpile::load(supply_dir);
```

dr-orlovsky:
Can you please update the PR to include the merged version?

astral-bitlight:
由于之前的版本不够完善，我修改了 import 的逻辑以便完整导入 articles 和 pile 的相关数据，请您看下是否符合 RGB 的设计。
这个版本要求：
1. 公开 FileSupply::FILENAME_ARTICLES 以便拼接 articles_path
2. 增加 stockpile.merge_articles 方法
3. 增加 filepile.export 方法
4. 增加 Stockpile::new 方法

dr-orlovsky (commented):
Why do we need to import articles from an existing contract?

The only case when we need articles is where we issue new contracts (tokens). The issuer MUST not trust to other issuers and must pick an issuer object signed by some developer - not from other contract via articles.

Articles is not something which can be separated from a contract. I do not get why we need to import them in the first place...

PR #300, opened 2025-03-13
0.11.1-alpha.2

zoedberg:


src/containers/anchors.rs
```diff
@@ -56,18 +59,36 @@ pub enum AnchoredBundleMismatch {
...
     }
 
     pub fn witness_id(&self) -> Txid { self.public.to_witness_id() }
+
+    /// Merges two [`SealWitness`]es keeping revealed data.
+    pub fn merge_reveal(&mut self, other: &Self) -> Result<(), SealWitnessMergeError> {
+        if self.dbc_proof != other.dbc_proof {
+            return Err(SealWitnessMergeError::DbcMismatch);
+        }
+        self.public.clone().merge_reveal(&other.public)?;
```
astral-bitlight, line 82: why `clone()` here?

invoice/src/parse.rs
```diff
@@ -260,22 +255,42 @@ impl FromStr for XChainNet<Beneficiary> {
...
+
+        let internal_pk = match internal_pk {
+            None => None,
+            Some(i) => {
+                if i.is_empty() {
+                    return Err(InvoiceParseError::Beneficiary(s!("missing internal pk")));
+                }
+                Some(
+                    InternalPk::from_str(i)
+                        .map_err(|_| InvoiceParseError::Beneficiary(s!("invalid internal pk")))?,
```
astral-bitlight, line 282:
```suggestion
                    InvoiceParseError::Beneficiary(format!("invalid internal pk: {e}"))
```

astral-bitlight, line 282: how about
```
                Some(InternalPk::from_str(i).map_err(|e| {
                    InvoiceParseError::Beneficiary(format!("invalid internal pk: {e}"))
                })?)
```

zoedberg: That's what I initially did, but then tests showed that `e` is an empty string, so this would become `invalid internal pk: `. Moreover I think the message is informative enough like it is now.

src/containers/anchors.rs
zoedberg: Good catch! It was an oversight, I will fix it

dr-orlovsky:
Will be reviewed in https://github.com/RGB-WG/rgb-std/pull/311

PR #301, opened 2025-03-18
feat(pile): add export method to FilePile

astral-bitlight:
We added this method so that we can import hoard + cache + keep + index data when importing a external contract.


dr-orlovsky (commented):
What is a rationale or a usecase for exporting Pile alone, without Stock (i.e. Plile and not Mound)?

BTW exporting in this way is computationally-expensive and slow

PR #302, opened 2025-03-21
Add purge contract functionality and error handling

astral-bitlight:


src/mound.rs
```diff
@@ -408,5 +417,45 @@ pub mod file {
...
+                .articles()
+                .contract
+                .meta
+                .name
+                .to_string();
+
+            self.contracts.remove(&contract_id);
+
+            if !keep_files {
+                let mut contract_dir = self.path().join(&contract_name);
```
dr-orlovsky, line 450: A small improvement

```suggestion
            let contract = self.contracts.remove(&contract_id) else {
                return Err(PurgeError::UnknownContract(contract_id));
            };

            if !keep_files {
                let contract_name = contract
                    .stock()
                    .articles()
                    .contract
                    .meta
                    .name
                    .to_string();
            
                let mut contract_dir = self.path().join(&contract_name);
```

dr-orlovsky, line 450: A small improvement

```suggestion
            let contract = self.contracts.remove(&contract_id) else {
                return Err(PurgeError::UnknownContract(contract_id));
            };

            if !keep_files {
                let contract_name = contract
                    .stock()
                    .articles()
                    .contract
                    .meta
                    .name
                    .to_string();
            
                let mut contract_dir = self.path().join(&contract_name);
```

dr-orlovsky (changes requested):
Can you please:
1) add to Mound outside of `fs` feature gate method named `remove_contract` moving the main logic there (without removing the directory)
2) change the existing method `purge_contract` to call that `remove_contract` and then remove the dir. The flag `keep_files` will become unnecessary.

astral-bitlight:
I've moved the core contract removal logic to remove_contract in the main Mound impl per your request. Kindly review.

dr-orlovsky:
Closing as not anymore applicable after v0.12 RC1.

Will add contract removal separately in one of future releases.

PR #303, opened 2025-03-21
feat(mound): Add export and import functions for Mound and FilePile

astral-bitlight:
[THIS IS A DRAFT, TO BE DONE...]

### DIFF
--- a/src/pile.rs
+++ b/src/pile.rs
@@ -296,16 +296,12 @@ pub mod fs {
             let path = path.as_ref();
             create_dir_all(path)?;
 
-            // 导出 hoard
             self.hoard.export(path.join("hoard"))?;
 
-            // 导出 cache
             self.cache.export(path.join("cache"))?;
 
-            // 导出 keep
             self.keep.export(path.join("keep"))?;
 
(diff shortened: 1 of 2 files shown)


src/pile.rs
```diff
@@ -284,4 +284,44 @@ pub mod fs {
...
+            self.cache.export(path.join("cache"))?;
+
+            self.keep.export(path.join("keep"))?;
+
+            self.index.save()?;
+
+            Ok(())
+        }
+
+        pub fn strict_encode<W: WriteRaw>(&self, writer: &mut StrictWriter<W>) -> io::Result<()> {
```
dr-orlovsky, line 310: It's better to avoid naming functions the same way as the trait methods. As an alternative, try to either derive `StrictEncode`/`Decode` directly on `FilePile` - or manually.

astral-bitlight: Okay, to avoid manual encoding, I've decided to derive these macros as you suggested. Here's a prerequisite PR (https://github.com/AluVM/sonic/pull/12) that adds encode/decode support to Aura.

PR #305, opened 2025-04-15
feat(invoice): add new invoice creation function and dependencies

astral-bitlight:


astral-bitlight:
Since RgbInvoice is currently a type alias for CallRequest, we can’t directly implement methods like new() on it. To address this, I propose refactoring it into a newtype wrapper (a zero-cost abstraction) with an explicit constructor

invoice/src/lib.rs
```diff
@@ -124,3 +126,46 @@ pub enum ParseInvoiceError {
...
+        beneficiary: RgbBeneficiary,
+        // Core parameters
+        value: Option<u64>,
+        expiry_time: Option<DateTime<Utc>>,
+        // Additional fields
+        api: Option<TypeName>,
+        call: Option<CallState>,
+        lock: Option<TinyBlob>,
+        endpoints: Option<impl Into<ConfinedVec<Endpoint, 0, 10>>>,
+    ) -> Self {
```
dr-orlovsky, line 144: This is why I prefer (and rust guidelines) recommend builders in such cases: calls to functions with too many unnamed arguments (some of which will be just `None`s) hard to read in the code. Clippy also grumbles about this.

Why you are not satisfied with the builder we have for the invoices?

invoice/src/lib.rs
```diff
@@ -124,3 +126,46 @@ pub enum ParseInvoiceError {
...
+        scope: T,
+        beneficiary: RgbBeneficiary,
+        // Core parameters
+        value: Option<u64>,
+        expiry_time: Option<DateTime<Utc>>,
+        // Additional fields
+        api: Option<TypeName>,
+        call: Option<CallState>,
+        lock: Option<TinyBlob>,
+        endpoints: Option<impl Into<ConfinedVec<Endpoint, 0, 10>>>,
```
dr-orlovsky, line 143: this rather should be
```suggestion
        endpoints: impl IntoIterator<Endpoint>,
```

And the method should panic if too many endpoints are provided

dr-orlovsky (changes requested):
Can you please also fix the clippy lints?

invoice/src/lib.rs
astral-bitlight: I think using the builder is totally ok, which is why this PR was initially marked as a draft

invoice/src/lib.rs
```diff
@@ -124,3 +126,46 @@ pub enum ParseInvoiceError {
     #[from]
     Bp(bp::ParseWitnessOutError),
 }
+
+pub struct RgbInvoice<T: Display + FromStr>(CallRequest<T, RgbBeneficiary>);
```
dr-orlovsky, line 130: I think we still should have the default value here. Also we need derives

```suggestion
#[derive(Clone, Eq, PartialEq, Debug, From)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[cfg_attr(
    feature = "serde",
    serde(
        transparent, 
        bound(
            serialize = "T: serde::Serialize, A: serde::Serialize",
            deserialize = "T: serde::Deserialize<'de>, A: serde::Deserialize<'de>"
        )
    )
)]
pub struct RgbInvoice<T: Display + FromStr = CallScope<ContractQuery>>(CallRequest<T, RgbBeneficiary>);
```

astral-bitlight:
I've added serde support and fixed constructor param issue

PR #314, opened 2025-05-22
chore: allow select coin by outpoint

Crayon-Shin-chan-bitlightlabs:
For the sake of user experience, wallets generally allow users to choose a specific bound asset to spend.

we can impl ourself's CoinSelect


dr-orlovsky (approved):
ACK bebad2be9c0ac7634e785ee58651445cf6bdcebf

PR #319, opened 2025-06-10
Improve accounting for the witness status in the history

dr-orlovsky:
Right now owned state reports just a state of its own witness mining.

However, it is desirable to know the worst confirmation in the whole of the history leading to the state.

This is purely UI enhancement (since invalid state is anyway excluded due to rollback procedure).

### DIFF
--- a/src/contract.rs
+++ b/src/contract.rs
@@ -110,6 +110,7 @@ impl<Seal> Assignment<EitherSeal<Seal>> {
     }
 }
 
+/// Element of the contract owned state, carrying information about its confirmation status.
 #[derive(Clone, PartialEq, Eq, Debug)]
 #[cfg_attr(
     feature = "serde",
@@ -117,9 +118,23 @@ impl<Seal> Assignment<EitherSeal<Seal>> {
     serde(bound = "Seal: serde::Serialize + for<'d> serde::Deserialize<'d>")
 )]
 pub struct OwnedState<Seal> {
+    /// Operation output defining this element of owned state.
     pub addr: CellAddr,
+
+    /// State assignment.
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub assignment: Assignment<Seal>,
+
+    /// Status of the state: the maximal confirmation depth for the whole history of operations
+    /// leading to this status, since genesis.
+    ///
+    /// If any operation has known multiple witnesses (related to RBF'ed transactions or lightning
+    /// channels), the best confirmation is used for that specific operation (i.e., the deepest
+    /// mined).
+    ///
+    /// Formally, if $H$ is a set of all operations in the history between genesis and this state,
+    /// and $W_o$ is a set of all witnesses for operation $o \in H$, the status here is defined as
+    /// $max_{o \in H} min_{w \in W_o} status(w)$.
     pub status: WitnessStatus,
 }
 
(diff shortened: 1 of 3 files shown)


will-bitlightlabs:
I spent considerable time understanding your changes, If I understand correctly, the `status` field in `OwnedState` is designed to reflect the most reasonable status within the transaction chain history, rather than the direct witness status of the specific transaction itself. Please let me know if my understanding is accurate.

However, I've noticed that the test case is currently failing with unexpected results:

**Expected vs Actual:**
- `wlt_1`: Should have `data: 200` and appropriate transfer status, but currently shows `data: 600` with `status: Genesis`
- `wlt_2`: Has correct `data: 400` but incorrect `status: Genesis` instead of the expected transfer status

It appears that both wallets are showing `Genesis` status instead of reflecting the actual transfer states.

src/pile.rs
```diff
@@ -92,12 +81,45 @@ impl WitnessStatus {
...
+    pub fn best(self, other: Self) -> Self {
+        if self.is_better(other) {
+            self
+        } else {
+            other
+        }
+    }
+
+    pub fn worst(self, other: Self) -> Self {
+        if self.is_better(other) {
```
will-bitlightlabs, line 111: I noticed that the `worst` method implementation seems to be inverted and now behaves identically to the `best` method:

```rust
pub fn worst(self, other: Self) -> Self {
    if self.is_better(other) {
        self  // This returns the better one, same as best()
    } else {
        other
    }
}
```

Shouldn't it be:
```rust
pub fn worst(self, other: Self) -> Self {
    if self.is_better(other) {
        other  //  Return the worse one
    } else {
        self
    }
}
```

src/contract.rs
```diff
@@ -370,21 +385,21 @@ impl<S: Stock, P: Pile> Contract<S, P> {
...
-    /// Get mining status for a given operation.
+    /// Get the best mining status for a given operation ("best" means "the most deeply mined").
     fn witness_status(&self, opid: Opid) -> WitnessStatus {
         self.pile
             .op_witness_ids(opid)
             .map(|wid| self.pile.witness_status(wid))
-            .max()
-            .unwrap_or(WitnessStatus::Genesis)
+            // "best" means "the most deeply mined"
+            .fold(WitnessStatus::Genesis, |best, other| best.best(other))
```
will-bitlightlabs, line 394: In the `witness_status` method, I have a concern about using `WitnessStatus::Genesis` as the initial value:

```rust
fn witness_status(&self, opid: Opid) -> WitnessStatus {
    self.pile
        .op_witness_ids(opid)
        .map(|wid| self.pile.witness_status(wid))
        .fold(WitnessStatus::Genesis, |best, other| best.best(other))
        //    ^^^^^^^^^^^^^^^^^^^^ This is the best possible status
}
```

Since `Genesis` is the best possible status, if an operation has no witnesses, this method would return the best status instead of indicating the absence of witnesses. 

Should we use `WitnessStatus::Archived` as the initial value instead, so that operations without witnesses are properly marked as having the worst status?

Please confirm your intended behavior for this case.

dr-orlovsky: The only operations which do not have a witness are operations with no closed seals (owned inputs), i.e. genesis and state extensions. And they are yeah, genesis - not archived.

All other ops are forced to have at least one witness by the consensus validation - otherwise they would never be included into the contract history.

src/pile.rs
dr-orlovsky: Fixed in https://github.com/RGB-WG/rgb-std/commit/a84f7cd3ccf7efdea7c908e4a0925681d16e7591

will-bitlightlabs:
I found another issue in the `get_status` method. The initial value for the ancestor fold operation should be `WitnessStatus::Genesis` instead of `WitnessStatus::Archived`:

```rust
// Current (incorrect):
.fold(WitnessStatus::Archived, |worst, other| worst.worst(other))

// Should be:
.fold(WitnessStatus::Genesis, |worst, other| worst.worst(other))
```

**Reason**: Since `Archived` is already the worst possible status, using it as the initial value means `worst.worst(other)` will always return `Archived` regardless of the actual ancestor statuses.

This causes all operations to be marked as `Archived` even when they should have better statuses like `Mined`.

I've tested this locally and it fixes the issue.

dr-orlovsky:
Yes, another finding...

Fixed in https://github.com/RGB-WG/rgb-std/pull/319/commits/de2389cdeb381429ed9033d4ab7c17fe54f0e3da

Also made encoding to be the same as before not to break the persistence binary compatibility with RC2

will-bitlightlabs:
> Yes, another finding...
> 
> Fixed in [de2389c](https://github.com/RGB-WG/rgb-std/commit/de2389cdeb381429ed9033d4ab7c17fe54f0e3da)
> 
> Also made encoding to be the same as before not to break the persistence binary compatibility with RC2

Great news! The latest code passes rbf tests and works as expected! 🎉

Thank you for the quick fixes and detailed explanations. The implementation now correctly handles the witness status accounting in the history.

PR #322, opened 2025-06-30
chore: add predicate for contract witness_ids

Crayon-Shin-chan-bitlightlabs:
At this stage, we can temporarily filter out witnesses with `Mined` and `Genesis` status when updating the state.


dr-orlovsky:
The filter can be just applied to the result returned by `witness_id` function - it will be strictly the same like in this PR, except will not make the method signature such complex.

